PR #22, opened 2016-01-21
TCOMP-52-TypeMapping and I/O interface

bchen-talend:


francisu:
We have done most of this another way, so closing this one.

PR #23, opened 2016-01-25
Feature/tcomp 70 make apis for context value computations

sgandon:
I'd like some people from the Studio to try this implmentation and tell if it fits their requirements for **context** values.


ycbai:
I think it is a good idea which use "Value Evaluator" to handle context script in runtime.

PR #26, opened 2016-01-29
TCOMP-75: Need add repository value for the parameter if it is from

jzhao-talend:
repository
https://jira.talendforge.org/browse/TCOMP-75
Note:Add repository value for all properties of specific
ComponentProperties


francisu:
Thanks for the pull request. I want to talk about this with @sgandon. I think the concept of a "repository_value" is probably not something for the framework, but rather it's something in DI. However, we do want to define this key in only one place, and the components can't depend on DI stuff. Also, this seems to want to be a flag on the COmponentProperties rather than on each property. So maybe the key/value mechanism can be extended to allow this on ComponentProperties in addition to Property. I do think we need some standard mechanism where components can declare "engine-specific" things (di would be considered an "engine"). Hopefully we can figure this out in the next couple of days.

jzhao-talend:
@francisu 
Thanks for your reply.
I get the information from DI team. https://jira.talendforge.org/browse/TUP-4014
They hope we can set the "repository_value" in component side as before.
Maybe we need more discuses about this.

sgandon:
Hi guys, I am sorry but this is too DI specific and we want those component to be as much as possible agnostic of the client.
**repository_value** has no meaning for Dataprep or TIC.
All this code can be done in the Studio without modifying the Components API.

jzhao-talend:
@sgandon 
ok. got it.

PR #27, opened 2016-02-01
TCOMP-75: Need add repository value for the parameter if it is from repository

jzhao-talend:
Add repository tag for properties of ComponenetProperties object


sgandon:
please see comments on https://jira.talendforge.org/browse/TCOMP-75.

PR #31, opened 2016-02-22
Features/tcomp 62

sgandon:
I created a pull request so that we can review/discuss about the implementation.


components-api/src/main/java/org/talend/components/api/adaptor/Adaptor.java
```diff
@@ -0,0 +1,20 @@
+package org.talend.components.api.adaptor;
+
+import java.util.Date;
+
+/**
+ *
+ */
+public interface Adaptor {
```
sgandon: This is not very clear what it deal with, the name seems a little to vague and abstract and also refer to some famous OO Design pattern that do seems to fit here. This seems to represent the container or the **context** (which I would prefer) in witch the component is executed.

components-api/pom.xml
```diff
@@ -177,6 +180,11 @@
 			<version>1.3</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
sgandon: We should get rid off this once we move to  Java 8.

components-api/src/main/java/org/talend/components/api/component/runtime/BoundedSource.java
```diff
@@ -0,0 +1,65 @@
...
+ * ({@link BoundedReader#splitAtFraction}).
+ * </ul>
+ *
+ */
+public interface BoundedSource extends Source {
+  /**
+   * Splits the source into bundles of approximately {@code desiredBundleSizeBytes}.
+   */
+  public abstract List<? extends BoundedSource> splitIntoBundles(
+      long desiredBundleSizeBytes, Adaptor adaptor) throws Exception;
```
sgandon: I wonder if we should not specialize all these Exception to something like ComponentException so that we have control over them and provide an automatic mean of dealing with them in a REST environment.

components-api/src/main/java/org/talend/components/api/component/runtime/SourceOrSink.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Checks that this source is valid, before it can be used in a pipeline.
+     */
+    public ValidationResult validate(Adaptor adaptor);
+
+    /**
+     * Get the list of schema names available for this {@code Source}.
+     */
+    public List<NamedThing> getSchemaNames(Adaptor adaptor) throws ComponentException;
```
sgandon: Is this the abstraction for getting Salesforce Module name or DB Table/View name ?

components-api/src/main/java/org/talend/components/api/component/runtime/BoundedSource.java
```diff
@@ -0,0 +1,65 @@
...
+   * This estimate is in terms of external storage size, before any decompression or other
+   * processing done by the reader.
+   */
+  public abstract long getEstimatedSizeBytes(Adaptor adaptor) throws Exception;
+
+  /**
+   * Whether this source is known to produce key/value pairs sorted by lexicographic order on
+   * the bytes of the encoded key.
+   */
+  public abstract boolean producesSortedKeys(Adaptor adaptor) throws Exception;
```
sgandon: I don't understand what it this method used for ? can you explain please.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSource.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public boolean producesSortedKeys(Adaptor adaptor) throws Exception {
+        return false;
+    }
+
+    @Override
+    public BoundedReader createReader(Adaptor adaptor) throws IOException {
+        if (properties instanceof TSalesforceInputProperties)
```
sgandon: I think this check should be done in an overriden version of the **initialize(Adaptor, ComponentProperties)** method so that we are sure that when this object is called it was properly initialized and there no need to check it anywhere else.
This would also mean that the *initialize* method should handle errors either with a result or an exception.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
```diff
@@ -0,0 +1,243 @@
...
+            element.setType(SchemaElement.Type.DATETIME);
+            element.setPattern("\"yyyy-MM-dd\'T\'HH:mm:ss\'.000Z\'\""); //$NON-NLS-1$
+        } else if (type.equals("double")) { //$NON-NLS-1$
+            element.setType(SchemaElement.Type.DOUBLE);
+        } else if (type.equals("currency")) { //$NON-NLS-1$
+            element.setType(SchemaElement.Type.DECIMAL);
+        }
+        element.setNullable(field.getNillable());
+
+        if (element.getType() == SchemaElement.Type.STRING) {
```
sgandon: this is strand to assume that the element may by a String when there is not setType as a String before here. This would mean that this method expects the paramter element is of type String somehow.

components-api/src/main/java/org/talend/components/api/component/runtime/Writer.java
```diff
@@ -0,0 +1,64 @@
...
+     * <p>The unique id that is given to open should be used to ensure that the writer's output does
+     * not interfere with the output of other Writers, as a bundle may be executed many times for
+     * fault tolerance. See {@link Sink} for more information about bundle ids.
+     */
+    public abstract void open(String uId) throws Exception;
+
+    /**
+     * Called for each value in the bundle.
+     */
+    public abstract void write(Map object) throws Exception;
```
sgandon: Why is Map used as a type for object to be written ?
Should it be Avro Record or indexedrecord type  ?

components-api/src/main/java/org/talend/components/api/adaptor/Adaptor.java
francisu: This is the name we have been struggling with all alone. We have variously called this "container" (the current code has ComponentRuntimeContainer) "framework", "engine", and now I'm trying "adaptor". Actually, my intention of this is that it does pretty well fit the adaptor pattern: http://www.oodesign.com/adapter-pattern.html. And we will provide adaptor modules which make up the difference between our "Adaptor" interface and the target.

I am not sure about "Context" as that is a bit too passive for me (like its only purpose is to contain some state). "Container" would probably be fine.

components-api/src/main/java/org/talend/components/api/component/runtime/BoundedSource.java
francisu: Yes, certainly, I will make that change.

components-api/pom.xml
francisu: Yes of course.

components-api/src/main/java/org/talend/components/api/component/runtime/SourceOrSink.java
francisu: Yes.

components-api/src/main/java/org/talend/components/api/component/runtime/Writer.java
francisu: Yes (it should move to Avro), I have left it as Map in this branch to stay close to master. Ryan is doing the separate work to move to IndexedRecord.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
francisu: This is all going to get replaced by the stuff that ChenBin and Ryan have been working on. This is just copied from current master.

components-api/src/main/java/org/talend/components/api/adaptor/Adaptor.java
sgandon: But adaptor is used to convert one type of object to another type, I don't see how it fits here ???
I is usually used to mean that there is a conversion, like Ryan's current work on providing a mean to adapt native Input objects to avro IndexRecords. He did a presentation to me and I like the idea very much.

components-api/src/main/java/org/talend/components/api/component/runtime/Writer.java
sgandon: Followin the talk I have this morning with Ryan, this will be an Object. 
The consumer of the Object will then call some Registry to get a factory to convert the object to avro index record if necessary. I like the idea because output component may also decide to use the native object if they can handle it. So we must make it an object.

components-api/src/main/java/org/talend/components/api/adaptor/Adaptor.java
francisu: An adaptor is to adapt one object type to another, not necessarily convert. But I'm not that strong on it (at least not as strong as you are in opposing it). I would like to hear other's opinions. We do need a name for this.

components-api/src/main/java/org/talend/components/api/component/runtime/BoundedSource.java
francisu: Me either.

components-api/src/main/java/org/talend/components/api/component/runtime/Writer.java
francisu: Ummmmm, I would like to be convinced of that. Not that I can't be. Maybe I can talk with you and Ryan a little later today?

components-api/src/main/java/org/talend/components/api/component/runtime/BoundedSource.java
sgandon: I think it is much better not to throw anything at all here and everywhere there is a Exception thrown. Because Exception is a not a checked exception there is not reason why we should declared it. I have left there removed them all except those methods dealing with IO that should throw a check ed IOException. Since it the runtime I believe it is the best approach.

francisu: I noticed that you changed your mind about this. I still like ComponentException.

sgandon: * The problem is that ComponentException is a RuntimeException so the catch is not enforced therefore the is not need to declare it if it does not have to be caught.
* then I have marked all the methods that need to do some IO access with a *throws IOException* but other methods should never throw exceptions or they will be runtime exceptions this is more simple for everyone.

PR #33, opened 2016-03-11
unify the Dynamic usage

bchen-talend:
There are two method for Dynamic exist.
org.talend.daikon.avro.util.AvroUtils.isDynamic(Schema schema){
return schema.getType() == Type.BYTES && schema.getLogicalType().getName().equals("dynamic");
}

Talend6OutgoingSchemaEnforcer.isDynamic(Field f){
return "id_Dynamic".equals(f.getProp("talend6.column.talendType"));
}

But we'd better to keep one way, this pull request is unify to use the second way to fix some bugs


components-api/src/main/java/org/talend/components/api/component/runtime/RuntimeHelper.java
```diff
@@ -33,12 +34,17 @@ public static Schema resolveSchema(RuntimeContainer container, SourceOrSink ss,
...
+            for (String key : objectProps.keySet()) {
+                field.addProp(key, objectProps.get(key));
+            }
+            copyFieldList.add(field);
         }
 
         int dynamicIndex = 0;
         for (Schema.Field se : copyFieldList) {
-            if (AvroUtils.isDynamic(se.schema())) {
+            if (Talend6OutgoingSchemaEnforcer.isDynamic(se)) {
```
sgandon: the problem is the your are using Talend6 stuff in component and you should never do that.
the *Talend6OutgoingSchemaEnforcer* is going to be move to routine.system of the Studio by Ryan.

components-api/src/test/java/org/talend/components/api/component/runtime/RuntimeHelperTest.java
```diff
@@ -22,12 +23,10 @@ public void testResolveSchema() throws Exception {
         Schema designSchema = SchemaBuilder.record("design").fields()
                 .name("f1").type().stringType().noDefault()
                 .name("f2").type().stringType().noDefault()
-                .name("dynamic").type().bytesType().noDefault()
+                .name("dynamic").prop(Talend6SchemaConstants.TALEND6_COLUMN_TALEND_TYPE, Talend6OutgoingSchemaEnforcer.TALEND6_DYNAMIC_TYPE).type().bytesType().noDefault()
```
sgandon: same comment as above. NO studio specific code here.

bchen-talend:
Thanks @sgandon, I change the way, please refer to https://github.com/Talend/components/pull/35.
so I close this one then.

PR #34, opened 2016-03-16
Tcomp 69

jzhao-talend:
The changes list:
1) Add "Port" for ProxyProperties and fixed proxy doesn't work
2) Add "endpoint" and "httpChunked " connection
3) Reuse bulk api code which used in old framework
4) Fixed Bulk Query in new salesforceinput component
5) Fixed query condition doesn't work
6) Refactor the salesforce output and bulk properties
7) Fixed outputbulk runtime and fixed junit
8) Fixed bulkExec runtime and fixed junit
9)Add default value for property
10)Fixed tSalesforceGetUpdated and tSalesforceGetDeleted UI problem


components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSink.java
```diff
@@ -55,8 +56,8 @@ public ValidationResult validate(RuntimeContainer adaptor) {
      *
      * @return the properties
      */
-    public TSalesforceOutputProperties getSalesforceOutputProperties() {
-        return (TSalesforceOutputProperties) properties;
+    public ComponentProperties getSalesforceOutputProperties() {
```
sgandon: I don't agree you change the type of this method, or if you want to do it, then change the name of the method otherwise it does not mean anything.

sgandon:
Thank you very much  for this pull request.
I'll do a review an comment on the code.

components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceConnectionProperties.java
```diff
@@ -85,14 +89,23 @@ public SalesforceConnectionProperties(String name) {
...
 
     @Override
-    public void setupLayout() {
-        super.setupLayout();
+    public void setupProperties() {
+        super.setupProperties();
 
         loginType.setValue(LOGIN_BASIC);
+        endpoint.setDefaultValue(URL);
+        timeout.setDefaultValue("60000");
```
sgandon: I don't think the default value should be a String, timeout is an integer so please make it an int.
` timeout.setDefaultValue(60000);`

jzhao-talend: @sgandon 
Yes. it should be int type.
But the defaultValue type of "Property" is String type.
It need to be changed to a Object type ?
Or we need use `timeout.setValue(60000);`

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSink.java
jzhao-talend: @sgandon 
Yes. you are right. I would chagne the mothod to
`public ComponentProperties getComponentProperties() {...}`

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkQuryInputReader.java
```diff
@@ -0,0 +1,166 @@
...
+import org.talend.components.api.container.RuntimeContainer;
+import org.talend.components.salesforce.runtime.SalesforceBulkRuntime.BulkResultSet;
+import org.talend.components.salesforce.runtime.SalesforceBulkRuntime.BulkResult;
+import org.talend.components.salesforce.tsalesforceinput.TSalesforceInputProperties;
+
+import java.io.IOException;
+import java.util.NoSuchElementException;
+
+public class SalesforceBulkQuryInputReader extends SalesforceReader<IndexedRecord> {
+
```
francisu: This class name is misspelled.

francisu:
Thanks for this, I will finish the detailed review tomorrow (my time).

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
```diff
@@ -49,11 +52,11 @@
 
     protected static final String API_VERSION = "34.0";
 
-    protected SalesforceProvideConnectionProperties properties;
+    protected ComponentProperties properties;
```
francisu: I don't understand why you removed the use of the SalesforceProvideConnectionProperties interface. This interface is important to provide the SalesforceConnectionProperties for any given Salesforce properties. How is this now handled?

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
```diff
@@ -162,13 +180,23 @@ public SessionRenewalHeader renewSession(ConnectorConfig connectorConfig) throws
             config.setConnectionTimeout(connProps.timeout.getIntValue());
         }
         config.setCompression(connProps.needCompression.getBooleanValue());
-        if (false) {
-            config.setTraceMessage(true);
-        }
+        //  config.setTraceMessage(connProps.httpTraceMessage.getBooleanValue());
+        config.setUseChunkedPost(connProps.httpChunked.getBooleanValue());
```
francisu: I prefer "if false" this way the statement actually compiles.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
```diff
@@ -280,4 +308,43 @@ protected Schema getSchema(PartnerConnection connection, String module) throws I
...
+            }
+            if(proxyPwd!=null && proxyPwd.length() > 0){
+                config.setProxyPassword(proxyPwd);
+            }
+        }else{
+            //No proxy.
+        }
+    }
+
 }
```
francisu: Most of what's in this setProxy method should shared code in components-common somewhere. Perhaps a static method in ProxyProperties, or something like ProxyPropertiesRuntimeHelper.

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforceoutput/TSalesforceOutputDefinition.java
```diff
@@ -68,7 +68,7 @@ public String getPartitioning() {
     @Override
     public Class<? extends ComponentProperties>[] getNestedCompatibleComponentPropertiesClass() {
         return concatPropertiesClasses(super.getNestedCompatibleComponentPropertiesClass(),
-                new Class[]{SalesforceModuleProperties.class});
+                new Class[] { SalesforceModuleProperties.class });
     }
```
francisu: This is non-standard formatting. Can you check your formatting (this applies to many other files).

components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceConnectionProperties.java
```diff
@@ -33,6 +33,8 @@
 
     public static final String OAUTH_URL = "https://login.salesforce.com/services/oauth2";
 
+    public Property endpoint = (Property)newString("endpoint").setRequired().setDefaultValue(URL);
+
     public static final String FORM_WIZARD = "Wizard";
```
francisu: The endpoint property should not be specified; it's always the same.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkExecReader.java
```diff
@@ -0,0 +1,153 @@
...
+            bulkConnection = ((SalesforceSource) getCurrentSource()).connect().bulkConnection;
+        }
+        return bulkConnection;
+    }
+
+    private Schema getSchema() throws IOException {
+        if (null == schema) {
+            schema = new Schema.Parser().parse(sprops.module.schema.schema.getStringValue());
+            //  querySchema = RuntimeHelper.resolveSchema(adaptor, getCurrentSource(), querySchema);
+        }
```
francisu: Why not resolveSchema?

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkFileWriter.java
```diff
@@ -0,0 +1,310 @@
...
+
+    private QuoteStatus quotestatus = QuoteStatus.AUTO;
+
+    /**
+     * DOC jzhao SalesforceBulkFileWriter constructor comment.
+     *
+     * @param salesforceWriteOperation
+     * @param adaptor
+     */
+    public SalesforceBulkFileWriter(SalesforceWriteOperation salesforceWriteOperation, RuntimeContainer adaptor) {
```
francisu: Please remove these DOC comments that are not filled in.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkFileWriter.java
```diff
@@ -0,0 +1,310 @@
...
+
+    }
+
+    @Override
+    public void open(String uId) throws IOException {
+        this.uId = uId;
+
+//        Schema schema = RuntimeHelper.resolveSchema(adaptor, sink,
+//                new Schema.Parser().parse(sprops.schema.schema.getStringValue()));
+
```
francisu: Why not resolveSchema?

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkFileWriter.java
```diff
@@ -0,0 +1,310 @@
...
+        if(lineEnd!=null) {
+            sb.append(lineEnd);
+            pw.write(sb.toString());
+        } else {
+            pw.println(sb.toString());
+        }
+
+    }
+
```
francisu: These methods above seem like general CSV handling methods having to do with escaping things. We should use some common code to do this (like a CSV library of some sort) - we should not have this in our implementation.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkQuryInputReader.java
```diff
@@ -0,0 +1,166 @@
...
+            bulkConnection = ((SalesforceSource) getCurrentSource()).connect().bulkConnection;
+        }
+        return bulkConnection;
+    }
+
+    private Schema getSchema() throws IOException {
+        if (null == querySchema) {
+            querySchema = new Schema.Parser().parse(properties.module.schema.schema.getStringValue());
+            //  querySchema = RuntimeHelper.resolveSchema(adaptor, getCurrentSource(), querySchema);
+        }
```
francisu: resolveSChema?

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkQuryInputReader.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    private BulkResultAdapterFactory getFactory() throws IOException {
+        if (null == factory) {
+            factory = new BulkResultAdapterFactory();
+            factory.setSchema(getSchema());
+        }
+        return factory;
+    }
+
```
francisu: This factory code is duplicated at least twice. You can put it in SalesforceReader as getBulkFactory() for example. No duplicated code please.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkQuryInputReader.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+        return true;
+    }
+
+    public BulkResult getCurrentRecord() throws NoSuchElementException {
+        return currentRecord;
+    }
+
+    // FIXME some duplicate code
+    protected void executeSalesforceBulkQuery() throws IOException, ConnectionException {
```
francisu: Please resolve this duplicated code before we merge. Otherwise it will probably never get resolved.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkQuryInputReader.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Override
+    public IndexedRecord getCurrent() {
+        try {
+            return getFactory().convertToAvro(getCurrentRecord());
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
```
francisu: Duplicated I think.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkQuryInputReader.java
```diff
@@ -0,0 +1,166 @@
...
+    @Override
+    public void close() throws IOException {
+        // No resources to close.
+    }
+
```
francisu: Remove this method if its empty

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
```diff
@@ -0,0 +1,598 @@
...
+            // boolean success = Boolean.valueOf(resultInfo.get("Success"));
+            // boolean created = Boolean.valueOf(resultInfo.get("Created"));
+            // String id = resultInfo.get("Id");
+            // String error = resultInfo.get("Error");
+            // if (success && created) {
+            // System.out.println("Created row with id " + id);
+            // } else if (!success) {
+            // System.out.println("Failed with error: " + error);
+            // }
+        }
```
francisu: What's this commented out code?

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
```diff
@@ -0,0 +1,598 @@
...
+        batchInfoList.add(info);
+        // For TDI-27909
+        closeJob();
+    }
+
+//    public Set<String> getQueryResultIDs() {
+//        return queryResultIDs;
+//    }
+
```
francisu: Commented code

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
```diff
@@ -0,0 +1,598 @@
...
+                return getQueryResultStream(jobID, batchID, resultID);
+            }
+            throw sfException;
+        }
+    }
+
+    private void renewSession() {
+        //  FIXME need to finish it.
+    }
+
```
francisu: Hook to stuff in SalesforceSourceOrSInk

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkQuryInputReader.java
jzhao-talend: Got it.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
jzhao-talend: @francisu 
I change it because tSalesforceOutputBulk no need to provide SalesforceConnectionProperties .
OK. I would recheck this. And try to separate the code.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
jzhao-talend: @francisu 
yes. this is wrong cover when I check the runtime code.
I would change back.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
jzhao-talend: @francisu 
ok. I would try to add a common one.

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforceoutput/TSalesforceOutputDefinition.java
jzhao-talend: @francisu 
ok. thanks for reminding me.

components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceConnectionProperties.java
jzhao-talend: @francisu 
I remember that we support Sandbox endpoint

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkExecReader.java
jzhao-talend: @francisu 
I think it should not always call this method.
Seems this mothod is used to generate dynamic field.
Because Chen Bin is help to fix some problem in https://jira.talendforge.org/browse/TCOMP-87
So I add a comment about adding a method to check the design schema include dynamic field or not.

components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceConnectionProperties.java
sgandon: yes please change the type of the setDefaultValue from String to Object. Francis and I agree on that.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkExecReader.java
```diff
@@ -0,0 +1,153 @@
+// ============================================================================
```
sgandon: This should not be a Reader but a Writer, this class purpose is to take the bundle csv file and output it to salesforce, this a sink/writer component not a reader.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkFileWriter.java
```diff
@@ -0,0 +1,310 @@
+// ============================================================================
```
sgandon: This could be generalized as a AvroBulkFileWrtier, there is nothing related to salesforce here.
Please move this to the module components-common.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkFileWriter.java
sgandon: To make this generic for Avro, I suggest you update the constructor to
```
BulkFileProperty bulkProperty,  RuntimeContainer adaptor) {
    public SalesforceBulkFileWriter(WriteOperation<WriterResult> writeOperation, Schema schema, String fileName, boolean append, boolean ignorNull, RuntimeContainer adaptor) {

```
or ideally this
```
public SalesforceBulkFileWriter(WriteOperation<WriterResult> writeOperation, Schema schema, BulkFileProperty bulkProperty,  RuntimeContainer adaptor)
```
where the you have a general property *BulkFileProperty* that contains all the properties and can be reusable by other components.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkExecReader.java
jzhao-talend: @sgandon 
> this class purpose is to take the bundle csv file and output it to salesforce

But this output action return a ResultSet , we also need to extract the result of the output action.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkFileWriter.java
jzhao-talend: @sgandon 
Yes. you are right.

francisu: But not "Avro", just BulkFileWriter. Everything is going to support Avro from now on (since that's the new standard for schemas), so we don't need that in the name.

sgandon: I agree just simply *BulkFileWrtier* is fine.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkExecReader.java
sgandon: Why should it return anything this is the final component that output to Salesforce.
Do you mean you also want to output something so that another component can be linked behind it ? We have not yet specified how to that with the component framework.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkExecReader.java
```diff
@@ -0,0 +1,153 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+final class SalesforceBulkExecReader extends SalesforceReader {
+
+
+    private BulkConnection bulkConnection;
+
+    protected SalesforceBulkRuntime bulkUtil;
```
sgandon: why not call it *bulkRuntime* instead of *bulkUtil* ? this would be more meaningful.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
```diff
@@ -0,0 +1,598 @@
...
+import java.io.InputStreamReader;
+import java.util.*;
+
+import com.sforce.async.*;
+import com.sforce.soap.partner.sobject.SObject;
+import com.sforce.ws.ConnectionException;
+import org.talend.components.salesforce.SalesforceOutputProperties;
+
+/*
+* Reuse the code in talend lib salesforceBulkAPI.jar
```
sgandon: This requires more java doc.
Something like.
Takes a single csv input file and split it into multiple files that are going to be sent to salesforce as buld data. (if I understood well).

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
```diff
@@ -0,0 +1,598 @@
...
+        BulkResult resultInfo;
+        BatchInfo b = batchInfoList.get(batchNum);
+        CSVReader rdr = new CSVReader(getBatchResultStream(job.getId(), b.getId()));
+
+        List<String> resultHeader = rdr.nextRecord();
+        int resultCols = resultHeader.size();
+        List<String> row;
+        while ((row = rdr.nextRecord()) != null) {
+            resultInfo = new BulkResult();
+            resultInfo.copyValues(getBaseFileRow());
```
sgandon: I don't understand here. It seems you are creating a bulk result with data that come from the original input file and data that come from the Salesforce batch result stream. Why ?

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkExecReader.java
jzhao-talend: @sgandon 
Yes.
In the old framework, we need to extract the bulk load ResultSet.
And link to a output component with main/reject line.
It can work independently or as a input component.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkFileWriter.java
jzhao-talend: got it.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkExecReader.java
jzhao-talend: @sgandon 
Yes. **bulkRuntime**  is better.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
jzhao-talend: @sgandon 
Yes. there should be more doc

> Takes a single csv input file and split it into multiple files that are going to be sent to salesforce as buld data

The bulk operation is not only "insert/update/upsert/delete"
It also include "query". And we support bulk query in input component.

I would be add more doc refer to:
https://resources.docs.salesforce.com/sfdc/pdf/api_asynch.pdf

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkFileWriter.java
jzhao-talend: Thanks for your suggestion. I would check this.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkFileWriter.java
jzhao-talend: @francisu 
OK. I would try to find one.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkQuryInputReader.java
jzhao-talend: ok. got it.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
jzhao-talend: @francisu 
I would clean the code of this class

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
jzhao-talend: @sgandon 
This for tSalesforceBulkExec as a input component.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkExecReader.java
sgandon: ok I understand now.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
sgandon: This is still not clear, can you explain more precisely, what is exactly output from the tSalesforceBulkExec component ?
Thanks.

jzhao-talend: tsalesforcebulkexec_simple_job
tsalesforcebulkexec_mainline_default_schema
tsalesforcebulkexec_rejectline_default_schema
Please see the screenshot.
1)If the record was loaded successfully, the **Main ** output would show the SObject id and other field of object(I just show the default and necessary field in the screenshot)
2)If the record can't be loaded, the **Reject ** output would showed the reason of the fail with the "error" field.

sgandon: Ok I got it but we don't handle **reject** right now.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
jzhao-talend: @francisu 
After investigate other components, I found this usage special code for salesforce.
So I didn't move this method to common.
I would move those code to common
`
java.util.Properties props = System.getProperties();
props.put("socksProxyHost",connProps.proxy.host.getStringValue());
props.put("socksProxyPort",connProps.proxy.port.getStringValue());
props.put("java.net.socks.username",connProps.proxy.userPassword.userId.getStringValue());
props.put("java.net.socks.password", connProps.proxy.userPassword.password.getStringValue());
`
Because we have a tSetProxy componet

sgandon:
Hi guys, what is the status of this PR ?

jzhao-talend:
@sgandon 
I have fixed most of the code.
I create a jira to fix a issue which I didn't solve
https://jira.talendforge.org/browse/TCOMP-135
Just because I still not too clear about the AvroRegistry.

You can check my new commits new.
Thanks.

components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceConnectionProperties.java
```diff
@@ -91,7 +93,7 @@ public void setupLayout() {
         super.setupLayout();
 
         loginType.setValue(LOGIN_BASIC);
-        endpoint.setValue(URL);
+        endpoint.setValue("\""+URL+"\"");
```
sgandon: I am sorry but you cannot set default values with quoted  (") strings in the component. The quotes are too specific for DI.

components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceGetDeletedUpdatedProperties.java
```diff
@@ -12,62 +12,34 @@
...
 
     public SalesforceGetDeletedUpdatedProperties(String name) {
         super(name);
     }
 
-    @Override
     public void setupProperties() {
         super.setupProperties();
-        module.connection = connection;
+        startDate.setValue("");
```
sgandon: I don't understand why these date values are set to empty string ?

sgandon:
I am ok with this, just one question though on the reason why the date properties get initialized with empty strings ?

components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceGetDeletedUpdatedProperties.java
jzhao-talend: @sgandon 
Because the value can't be NULL, it should be set a initialize value.
Set the value to empty means that would notice customer to specify a value for those two properties.

jzhao-talend: Hi Sébastien,
Yes.
This should be done by studio side. And they have been fixed string value with double quotes
So I have change back in this commit:
https://github.com/Talend/components/commit/db35dede8c63f57624d4829f235138dda22d10a0

Regards.

发件人: sgandon [mailto:notifications@github.com]
发送时间: Tuesday, March 29, 2016 3:26 AM
收件人: Talend/components <components@noreply.github.com>
抄送: Jin Zhao <jzhao@talend.com>
主题: Re: [Talend/components] Tcomp 69 (#34)


In components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceGetDeletedUpdatedProperties.java<https://github.com/Talend/components/pull/34#discussion_r57618808>:

>      public void setupProperties() {

>          super.setupProperties();

> -        module.connection = connection;

> +        startDate.setValue("");

I don't understand why these date values are set to empty string ?

sgandon: ok, I guessed you did not want it to be null,  so why can't it be null ? There is not reason from the component side not to be null.

jzhao-talend: @sgandon 
Sorry for not clear explanation. I change it to empty string.
Just because with the null value those two component can't be load.
I got NPE in "DateController.java" when I tried to create a component.
Do you think this should be solved by Studio team ?

sgandon: Yes I do think so. The Studio should be able to cope with default values set to null.

jzhao-talend: @sgandon 
ok. I would remove set the empty value for Date type.
And create a issue for studio team.

sgandon:
Let me know when you are done so that I merge the PR to master.

PR #36, opened 2016-03-21
TCOMP-117  : create virtual component definition

sgandon:


sgandon:
this has been done in another PR.

PR #37, opened 2016-03-23
TCOMP-87 dynamic type2

bchen-talend:
I'm not sure if this PR will be accepted, but anyway I just put what I have done here and you all can decide it, tell me any change I need to do :)

@RyanSkraba I move the schema resolution into validate method, but about getExtendedFields(), I'm not very sure, appreciate if you can help on this, feel free commit on this branch if you want :)

and about the return type of HasSchemaProperty, I don't change to Map<String, Schema> as I don't know how to set the key now.


RyanSkraba:
This looks good to me!  :+1:  It's pretty neat to see the evolution -- the code looks cleaner like this!

components-salesforce/src/test/java/org/talend/components/salesforce/SalesforceTestBase.java
```diff
@@ -156,6 +156,18 @@ protected void setupModule(SalesforceModuleProperties moduleProps, String module
         moduleProps = (SalesforceModuleProperties) checkAndAfter(f, "moduleName", moduleProps);
     }
 
+    protected void setupModuleWithEmptySchema(SalesforceModuleProperties moduleProps, String module) throws Throwable {
```
RyanSkraba: Does this need to be updated since empty schema is no longer the trigger for dynamic?

PR #38, opened 2016-03-24
TCOMP-117:  Add interface of virtual component

sgandon:
https://jira.talendforge.org/browse/TCOMP-117


sgandon:
I have removed the connection type from the ComponentDefinition as it is too specific to the Studio.
Let's assume for now that all components inside a virtual component will have an on_subjob_ok for now.
I have merged it manually.
Thanks for your contribution.

PR #39, opened 2016-03-30
1. complete built-in Avro type

bchen-talend:
2. fix a bug for Date pattern

### DIFF
--- a/components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceAvroRegistry.java
+++ b/components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceAvroRegistry.java
@@ -3,8 +3,6 @@ package org.talend.components.salesforce.runtime;
 import com.sforce.soap.partner.DescribeSObjectResult;
 import com.sforce.soap.partner.Field;
 import org.apache.avro.Schema;
-import org.apache.avro.SchemaBuilder;
-import org.apache.avro.SchemaBuilder.FieldAssembler;
 import org.talend.daikon.avro.AvroConverter;
 import org.talend.daikon.avro.AvroRegistry;
 import org.talend.daikon.avro.SchemaConstants;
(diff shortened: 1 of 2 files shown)


components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceAvroRegistry.java
```diff
@@ -119,31 +115,29 @@ private Schema inferSchemaField(Field field) {
...
+                break;
+            case _int:
+                base = AvroTypes._int();
+                break;
+            case currency:
+                base = AvroTypes._decimal();
+                break;
+            case date:
+                base = AvroTypes._date();
+                base.addProp(SchemaConstants.TALEND_COLUMN_PATTERN, "yyyy-MM-dd"); //$NON-NLS-1$
```
RyanSkraba: Can we put these two Salesforce specific date patterns someplace as a constant?

RyanSkraba:
Looks good to me!

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceAvroRegistry.java
bchen-talend: Yes, it will be best.
Another issue I would like to mention here and do next is, I think the pattern setting should not in here, I would like to add db_type even for salesforce, it's necessary. For example, if the user change the date column's type from id_Date to id_String and with pattern(dd-MM-yyyy), can't support it on the current implement.
I would like we have db_type which tell how to convert the data from String to java.util.Date, and then convert the value from java.util.Date to whatever the type user define, if it's still id_Date, means use ConvertDate.convertToAvro, if it's id_String with pattern dd-MM-yyyy, we should use other converter like DateToStringConveter(we don't have this kind of thing now, and I also would like to add later)

PR #40, opened 2016-03-31
TCOMP-126 add schema path to connectors

sgandon:
I have adde 2 services APIs : 

1. To query for available connectors for a given Component Properties,
 the implementation is done in the ComponentProperties but this could be moved to ComponentDefinition I think.
2. to query for Schema for a given connector and ComponentProperties.
I need a review on this please.


RyanSkraba:
For info, also implies the branch in daikon: https://github.com/Talend/daikon/tree/feature/TCOMP-126_add_schema_path_to_connectors

components-api/src/main/java/org/talend/components/api/component/InputConnector.java
```diff
@@ -12,31 +12,24 @@
 // ============================================================================
 package org.talend.components.api.component;
 
-public abstract class AbstractComponentConnection {
```
RyanSkraba: Delete me ;)

RyanSkraba:
Hello!

I didn't manage to finish the entire code review!  But I did all of the files up to components-salesforce and it Looks Good To Me for merging.

I didn't have any comments during the review.  I think we discussed this change quite a bit beforehand so nothing was too surprising!

One thing that we'll have to think about in the future is how Connectors and Components fit together to create an entire job graph -- it seemed clearer to me when Connectors lived in the ComponentDefinition.  But I'm not worried about that immediately, and it shouldn't stop you from merging this!!

francisu:
I would like to play with this a bit. I would like the connection type to go away. Connectors can simply be named. I have only looked at a little of this though and will try to get to it this weekend. I will make a branch off of this one with my suggestions. 

I'm concerned about creating anything like Rejectable as it's wiring a type of connection into the framework. I'm OK with some predefined names of things (like we do with Forms), so we could name predefined connection names of "MAIN" and "REJECT". But there should be nothing preventing a component from having multiple connections that reject things with different names.

Also, it looks like there are small differences in the code formatting (perhaps between intellij and the eclipse standard). We should try to figure this out before the TAB meetings in April.

jbonofre:
Just curious: does the connector has to deal with watermarks and timestamps (optionally for course). If it's not really required for connector dealing with bounded data, it could be interesting for unbounded data. Just thinking loud, not sure it makes sense to put this directly in the connector API.

francisu:
@jbonofre Yes, this is all part of the runtime API stuff. It's found here: https://github.com/Talend/components/tree/master/components-api/src/main/java/org/talend/components/api/component/runtime and you will note a more than passing resemblance to GDF.

jbonofre:
Thanks Francis !
FYI, yesterday we talked about the IO API. Basically, no deep change in the concepts, but some utilities are coming (especially to deal with timestamps, watermarks, really decouple the IO from the runners, ...).

sgandon:
@francisu there a real semantic difference between main and reject stream this is why we thought it was better to have a real type to mark this difference. Here is why they are different.

The idea behind that (as discussed with Ryan) is that the reject process is not done in the component but through some Schema enforcer component that may be plugged or not in-between components. Those schema enforcer would have the task of checking if the Indexed record coming out of the main(s) (which can be many) connector(s) match the expected schema. And if not, the schema enforcer will call back the Component runtime if it implements Rejectable (see https://github.com/Talend/components/pull/40/files#diff-169b25494b4e05158a6ee615a3e2de63 ) in order to create a Rejectable record, or if this interface is not implemented it schema enforcer will handle those errors by itselft. This notion of schema enforcer as a separate entity is already implemented in the Studio code generation engine.

RyanSkraba:
@francisu : knowing what a pain it is today to iterate through all of the DATA / FLOW_REF / MAIN tags that I still don't completely understand in the studio today, I wouldn't miss the connection type!

But we debated quite a bit about the different types of connectors and the Rejectable interface and went through a user story to try and define what we want -- there's two big issues:

a) The main/reject outputs are overwhelmingly the dominant pattern that we use in Talend components and should be easy to implement, and

b) Is it even possible to do what we want without having this behaviour "codified" in an interface?   In addition to the semantic difference that Sebastien mentions, won't it need to be treated differently in today's javajet?

It would be great if (b) were possible today, but I couldn't see how it would be possible.  Very open to ideas!

@jbonofre All of this would certainly be at the data exchange level in Beam, agreed?

francisu:
I merged master into this (and the corresponding Daikon branch) just to keep it up to date. That being said, I prefer the approach to connectors that was previously agreed and closer to the existing implementation. Based on that, most of these changes to the API don't make sense to me.

I think:

1. Connectors should not have a typed, but all be named. There would be any number of input or output connectors allowed. We can have reserved names for "Main" and "Reject", as we do with Forms.
2. There should be a maximum number of input or output connectors allowed.
3. You can ask the ComponentService to give you the Schema based on the connector name, ComponentProperties (and input/output). THis is close to what you have, but your new Connector is not a NamedThing. I would prefer to replace the Connector with just a String.
4.About the issue of Rejection - I don't see an implementation of this new interface, and I think we should consider modeling a "schema validation thing" as simply a transformation component (I realize we don't get have the transformer API). The designer UI can decide how to model this for the user, but it will make things simpler if all parts of a Flow are components of one type of another, rather than components and something else. You can imagine that the schema validation can do any number of validation sort of things, including correcting data, etc. All of which fits nicely in the transformation component. So I'm still opposed to a special interface for this.

I would not go forward with this branch as it is.

@sgandon let's talk about this further tomorrow if you like.

RyanSkraba:
I'd like to add my two cents :)  This is the specific story for running a component with a reject flow that we did not find an alternative for:

1. An input component (`tSalesforceInput`) has been created to read all records from the Accounts module.
2. The user modifies the schema so that it only includes Id and Name of type Integer (i.e., all accounts that do not have a name with exclusively digits should be on the REJECT flow).
3. At run-time, the tSalesforceInput does the query `SELECT Id, Name from Accounts` and emits the data with its **actual** data (long and String, respectively).
4. In the Studio, the `Talend6OutgoingSchemaEnforcer` receives the actual data and tries to coerce it to an integer so it can be stored in the POJO.
5. If there was a NumberFormatException by the enforcer, it needs to be taken off the MAIN flow and put on the REJECT flow.
6. The `tSalesforceInput` is the right place to decide how a failed MAIN output should appear on the REJECT output, since it is the one who decides the schema of the REJECT output.

The `Rejectable` interface was the proposition for the Studio to go back to the component and say "actually..."  I'm totally open to any other solution, because I think your points are *all* very valid!  At the time, we just didn't see any other way to do this...

Can you clarify your point #2 with an example?  (There should be a maximum number of input or output connectors allowed.)

The thing that **always** confuses me with component definition is where we define how many connections it *supports* (I can give you two outputs with this schema or that schema) versus the connections that it has been *configured with* in the job.  (row1 should have this schema and connect to this next component, row2 should have that schema and connect to that component).

sgandon:
About the number of connections, the API is dynamic and allow the component to decide which connection is available according to existing ones and according to the component properties value which may be the case some time. This is why we use a dynamic java method (`public Set<Connector> getAvailableConnectors(Set<Connector> existingConnectors, boolean isOutputConnection)`) to get the list of connector and why I don't think there is a need to specify them statically as it does not match all the requirements.

francisu:
@sgandon I agree with your comment above about having the connector be dynamic. I'm not sure about the Connector object itself though. So it seems the Connector object can represent both the possibility of a Connector and and actual connection that is made. In the latter case, it needs a name.

francisu:
@RyanSkraba regarding my point 2, it's probably not important to have a max input or output connectors.  It's probably enough just to have a list of Connector objects which are possible connectors and each has a name and whether it's input or output. I support @sgandon's idea that this should be dynamic.

francisu:
Seb and I spoke about this, and we agreed:

1. Remove the Rejectable interface
2. Add a NamedConnector class (making MAIN and REJECT defined names)
3. Remove the ConnectorType

PR #42, opened 2016-04-06
TCOMP-141: Change the virtual component interface and add the implement in salesforce

jzhao-talend:
https://jira.talendforge.org/browse/TCOMP-141
1)Change the interface of virtual component
2)Add implement in tSalesforceOutputBulkExec


nrousseau:
We might need to test / validate with ycbai first, to make sure everything is ok before merge to master

francisu:
I'm OK with merging this when you are ready.

PR #43, opened 2016-04-07
TDI-35445: Oracle input for Dataprep

wwang-talend:
https://jira.talendforge.org/browse/TDI-35445


components-oracle/readme.txt
```diff
@@ -0,0 +1,2 @@
+//install the ojdbc7.jar to your local repository firstly.
+mvn install:install-file -DgroupId=org.talend.libraries -DartifactId=ojdbc7 -Dversion=6.2.0 -Dfile=C:/Users/Talend/Desktop/ojdbc7.jar -Dpackaging=jar -DgeneratePom=true
```
sgandon: Do you really need that, if this jar is important you should upload it to talend nexus server.

francisu:
Thanks for this. I have some general comments to try to improve sharing. I like the structure of what you have done for reusability. I would remove the "CommonDefinition" class (as it does not appear to add much to AbstractComponentDefinition -- perhaps you can propose changes in AbstractComponentDefinition if you want new stuff there), and move all of the non-Oracle-specific classes into components-common (even though they might be specific only to databases, we want components-common to have anything that's shared between multiple components), as they will be shared between all database connections. So long as they don't depend on any additional Jar files (and they appear not to), then the should be fine there. 

I will make some more comments related to specific code.

components-oracle/src/main/java/org/talend/components/oracle/DBConnectionProperties.java
```diff
@@ -0,0 +1,131 @@
...
+import org.talend.daikon.properties.presentation.Widget;
+import org.talend.daikon.properties.presentation.Widget.WidgetType;
+
+public class DBConnectionProperties extends ComponentProperties implements DBProvideConnectionProperties, ComponentReferencePropertiesEnclosing {
+
+    // basic setting start
+    public Property               host                  = (Property) newString("host").setRequired(true);
+
+    public Property               port                  = (Property) newString("port").setRequired(true);
+
```
francisu: We should make a HostPortProperties and have these be in components-common

components-oracle/src/main/java/org/talend/components/oracle/DBConnectionProperties.java
```diff
@@ -0,0 +1,131 @@
...
+        mainForm.addRow(userPassword.getForm(Form.MAIN));
+
+        mainForm.addRow(jdbcparameter);
+
+        Form advancedForm = CommonUtils.addForm(this, Form.ADVANCED);
+        advancedForm.addRow(autocommit);
+
+        // only store it, will use it later
+        Form refForm = CommonUtils.addForm(this, Form.REFERENCE);
+
```
francisu: Do this the way SalesforceConnectionProperties works with the Form.REFERENCE

components-oracle/src/main/java/org/talend/components/oracle/DBConnectionProperties.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public void refreshLayout(Form form) {
+        super.refreshLayout(form);
+
+        String refComponentIdValue = getReferencedComponentId();
+        boolean useOtherConnection = refComponentIdValue != null
+                && refComponentIdValue.startsWith(getReferencedComponentName());
+        
+        if (form.getName().equals(Form.MAIN)) {
```
francisu: See SalesforceConnectionProperties.refreshLayout() for the REFERENCE stuff and do it the same way.

components-oracle/src/main/java/org/talend/components/oracle/DBInputProperties.java
```diff
@@ -0,0 +1,75 @@
...
+import org.talend.components.api.properties.HasSchemaProperty;
+import org.talend.components.common.SchemaProperties;
+import org.talend.daikon.properties.Property;
+import org.talend.daikon.properties.Property.Type;
+import org.talend.daikon.properties.presentation.Form;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class DBInputProperties extends DBCommonProperties implements HasSchemaProperty {
+
```
francisu: I think we want a shared set of properties for Schema/Table which is the same as Schema/Module for Salesforce. We already talked about this when I was last in China. We should have something shared that handles this (the only thing different is the labels of the Table (for db it's table, for Salesforce it's module). All of the other processing should be the same. We can use a subclass to handle those differences.

components-oracle/src/main/java/org/talend/components/oracle/runtime/DBReader.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.NoSuchElementException;
+
+import org.apache.avro.Schema;
+import org.apache.avro.generic.IndexedRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.talend.components.api.component.runtime.AbstractBoundedReader;
+import org.talend.components.api.component.runtime.BoundedSource;
+import org.talend.components.api.container.RuntimeContainer;
+import org.talend.components.oracle.DBInputProperties;
```
francisu: This looks like a generic database reader with JDBC (or whatever). This should go into a new module called "components-common-db" which can depend on the generic database libraries.

pteyssier:
Postpone to 6.3 certainly. I will update the label when we will want an integration (We will surely update this PR once more by then)

pteyssier:
JDBC integrated in DP, no reason to keep this PR opened

PR #45, opened 2016-04-11
TCOMP-146: Need to add back runtime code for new tSalesforceConnection

jzhao-talend:
https://jira.talendforge.org/browse/TCOMP-146

### DIFF
--- a/components-api/src/main/java/org/talend/components/api/component/RunnableComponentDefinition.java
+++ b/components-api/src/main/java/org/talend/components/api/component/RunnableComponentDefinition.java
@@ -1,0 +1,7 @@
+package org.talend.components.api.component;
+
+import org.talend.components.api.component.runtime.SourceOrSink;
+
+public interface RunnableComponentDefinition {
+    public SourceOrSink getRuntime();
+}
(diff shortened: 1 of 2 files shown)


sgandon:
I am ok with this pull request but I  wonder if it would be best to have another interface to have the validate and schema api in something more generic that SourceOrSink something that means is can be used both at runtime and design time whereas the SourceOrSink are more for getting the data or writing the data. Some kind of *Connector* or *SchemaProvider* or *Connection*. To me those would be used at design time and SourceOrSink at runtime. I am also wondering about the dependencies at design time and runtime. do we need to make that difference.

francisu:
I would support just adding getRuntime() at the ComponentDefinition level, the InputComponentDefinition and OutputComponentDefinition override this. I think the extra interface is not necessary and confusing.

@sgandon I don't think we need something different for runtime or design time, as what you do is essentially the same. In any case, that does not have to do this with particular issue, so we could discuss that separately. This is more of a bug fix.

francisu:
Replaced by similar pull request that uses EndpointComponentDefinition

PR #47, opened 2016-04-12
Tcomp 148 components rename

jzhao-talend:
Rename the name of new salesforce components.


sgandon:
I am checking it right now.

PR #50, opened 2016-04-19
TCOMP-128: Fixed runtime of tSalesforceGetUpdated and tSalesforceGetDeleted

jzhao-talend:
1)Fixed runtime of tSalesforceGetUpdated and tSalesforceGetDeleted
2)Add junit for the modification (only can test deleted record, I am not clear about the udpated records)
3)Because the network problem in china side. I only run the mvn install for "component-salesforce". it can be run successfully.


jzhao-talend:
Also a change for tdi-studio-se
https://github.com/Talend/tdi-studio-se/pull/341

PR #51, opened 2016-04-21
TDI-35834: Salesforce connection failed when job migrate to new framework

jzhao-talend:


components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
```diff
@@ -175,8 +176,12 @@ protected ConnectionHolder connect(RuntimeContainer container) throws IOExceptio
 
         ConnectorConfig config = new ConnectorConfig();
         config.setUsername(StringUtils.strip(connProps.userPassword.userId.getStringValue(), "\""));
-        config.setPassword(StringUtils.strip(connProps.userPassword.password.getStringValue(), "\"")
-                + StringUtils.strip(connProps.userPassword.securityKey.getStringValue(), "\""));
+        String password = StringUtils.strip(connProps.userPassword.password.getStringValue(), "\"");
```
jzhao-talend: This fixed the for wizard migration.
Old password value is set into new password field. So the securityKey is null.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
```diff
@@ -132,12 +132,13 @@ protected BulkConnection connectBulk(ConnectorConfig config) throws ComponentExc
 
     protected PartnerConnection doConnection(ConnectorConfig config) throws ConnectionException {
         SalesforceConnectionProperties connProps = properties.getConnectionProperties();
+        String endpoint = connProps.endpoint.getStringValue();
```
jzhao-talend: This fixed we didn't use the endpoint value in the component setting.

jzhao-talend:
About migration code in tdi-studio-se would let studio side to check it.
https://github.com/Talend/tdi-studio-se/pull/344

PR #52, opened 2016-04-21
TDI-35893 Move getName() getPngImagePath() Constructor(String compone…

gonchar-ivan:
Move getName() getPngImagePath() Constructor(String compone…ntName) to AbstractComponentDefinition class, because it seems to be common functionality

### DIFF
--- a/components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
+++ b/components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
@@ -14,7 +14,6 @@ package org.talend.components.api.component;
 
 import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
```diff
@@ -14,7 +14,6 @@
 
 import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
```
gonchar-ivan: Unused import

components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
```diff
@@ -25,10 +24,26 @@
 import org.talend.daikon.exception.TalendRuntimeException;
 
 public abstract class AbstractComponentDefinition extends AbstractTopLevelDefinition implements ComponentDefinition {
+    
+    /**
+     * Component name.
+     * It is used to define component name, which will be displayed in Studio Palette
+     * Also it is used by bnd library to create OSGi bundle
+     */
+    private String componentName;
```
gonchar-ivan: Moved componentName field and related methods from SalesforceDefinition class

components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceDefinition.java
```diff
@@ -13,42 +13,19 @@
...
-    public String getPngImagePath(ComponentImageType imageType) {
-        switch (imageType) {
-        case PALLETE_ICON_32X32:
-            return componentName + "_icon32.png"; //$NON-NLS-1$
-        }
-        return null;
-    }
-
-    @Override
-    public String getDisplayName() {
```
gonchar-ivan: There is getDisplayName() implementation in AbstractTopLevelDefinition class.
It seems Salesforce should use that implementation

PR #57, opened 2016-04-27
TCOMP-71: add url handler for different platform to deal with the authorize url

bchen-talend:


components-common-oauth/src/main/java/org/talend/components/common/oauth/DefaultOauthAuthorizeUrlHandler.java
```diff
@@ -0,0 +1,12 @@
+package org.talend.components.common.oauth;
+
+/**
+ * For component runtime, just display the url on console, and let the user paste it into browser
+ */
+public class DefaultOauthAuthorizeUrlHandler implements OauthAuthorizeUrlHandler {
+    @Override
+    public void display(String url) {
+        System.out.println("Paste this URL into a web browser to authorize Salesforce Access:");
```
sgandon: You are mentionning Salesforce here in a Common Oauth class. this you be removed.

bchen-talend: Yes, I need to remove salesforce here, and for the old component, this information will be print on the console and let the user copy/paste this url into browser. I don't know if there is a better way to do it.

sgandon:
Sorry, but I did not understand the purpose of this PR, now Nicolas explained it to me and this is not possible to resolve your problem this way.
You have to think of the Component being on a remote server side so you cannot display message to the user using some System.out (which you should never do anyway). You therefore cannot expect any interface of the Components to be implemented on the Client side (there are some rare exceptions for storage repository, but that where made on purpose).
I'll close this PR.

PR #58, opened 2016-04-27
TUP-4439: Fix UI issue for salesforce connection/input/output/...

jzhao-talend:


components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforceoutput/TSalesforceOutputProperties.java
```diff
@@ -75,15 +75,25 @@ public void refreshLayout(Form form) {
...
+                Schema s = SchemaBuilder.record("Main")
+                        .prop(Talend6SchemaConstants.TALEND6_IS_READ_ONLY, "true")//$NON-NLS-1$
+                        .fields().name("salesforce_id")
+                        .prop(Talend6SchemaConstants.TALEND6_COLUMN_CUSTOM, "true")//$NON-NLS-1$
+                        .prop(Talend6SchemaConstants.TALEND6_IS_READ_ONLY, "false")//$NON-NLS-1$
+                        .prop(Talend6SchemaConstants.TALEND6_COLUMN_TALEND_TYPE, "id_String")//$NON-NLS-1$
+                        .prop(Talend6SchemaConstants.TALEND6_COLUMN_LENGTH, "255")//$NON-NLS-1$
+                        .type().stringType().noDefault().endRecord();
+                module.main.schema.setValue(s);
             }
```
francisu: We need to fix this. Components should never use Talend6SchemaConstants - that's DI only. Only SchemaConstants are allowed.

PR #61, opened 2016-04-28
TUP-4439: Change outh "clientSecret" type  and adjust output default reject schema field

jzhao-talend:
1) Change outh "clientSecret" type to a password
2) Adjust output components default reject schema field refer old components


jzhao-talend:
About Talend6SchemaConstants problem, seems we have a jira to fix it.
https://jira.talendforge.org/browse/TCOMP-176
So no change over here

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforcebulkexec/TSalesforceBulkExecProperties.java
```diff
@@ -75,4 +79,18 @@ public void setupProperties() {
...
+                .prop(Talend6SchemaConstants.TALEND6_IS_READ_ONLY, "true")//$NON-NLS-1$
+                .fields().name("error")//$NON-NLS-1$
+                .prop(Talend6SchemaConstants.TALEND6_COLUMN_CUSTOM, "true")//$NON-NLS-1$
+                .prop(Talend6SchemaConstants.TALEND6_IS_READ_ONLY, "false")//$NON-NLS-1$
+                .prop(Talend6SchemaConstants.TALEND6_COLUMN_TALEND_TYPE, "id_String")//$NON-NLS-1$
+                .prop(Talend6SchemaConstants.TALEND6_COLUMN_LENGTH, "255")//$NON-NLS-1$
+                .type().stringType().noDefault().endRecord();
+        schemaReject.schema.setValue(s);
+    }
+
```
francisu: Should not use Talend6SchemaConstants - only SchemaConstants.

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforcegetservertimestamp/TSalesforceGetServerTimestampProperties.java
```diff
@@ -45,9 +47,13 @@ public TSalesforceGetServerTimestampProperties(String name) {
...
-        date.addProp(SchemaConstants.TALEND_COLUMN_PATTERN, "yyyy-MM-dd'T'HH:mm:ss'.000Z'");
-        Schema s = SchemaBuilder.record("Root").fields().name("ServerTimeStamp").type(date).noDefault().endRecord();
+        Schema s = SchemaBuilder.record("Main")
+                .prop(Talend6SchemaConstants.TALEND6_IS_READ_ONLY, "true")//$NON-NLS-1$
+                .fields().name("ServerTimeStamp")
+                .prop(Talend6SchemaConstants.TALEND6_COLUMN_TALEND_TYPE, "id_Date")//$NON-NLS-1$
+                .prop(SchemaConstants.TALEND_COLUMN_PATTERN, "yyyy-MM-dd'T'HH:mm:ss'.000Z'")
+                .prop(Talend6SchemaConstants.TALEND6_COLUMN_LENGTH, "20")//$NON-NLS-1$
+                .type(AvroTypes._date()).noDefault().endRecord();
         schema.schema.setValue(s);
```
francisu: No Talend6 stuff.

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforceoutput/TSalesforceOutputProperties.java
```diff
@@ -100,4 +102,29 @@ public void refreshLayout(Form form) {
...
+                .prop(Talend6SchemaConstants.TALEND6_COLUMN_LENGTH, "255")//$NON-NLS-1$
+                .type().stringType().noDefault().name("errorMessage")//$NON-NLS-1$
+                .prop(Talend6SchemaConstants.TALEND6_COLUMN_CUSTOM, "true")//$NON-NLS-1$
+                .prop(Talend6SchemaConstants.TALEND6_IS_READ_ONLY, "false")//$NON-NLS-1$
+                .prop(Talend6SchemaConstants.TALEND6_COLUMN_TALEND_TYPE, "id_String")//$NON-NLS-1$
+                .prop(Talend6SchemaConstants.TALEND6_COLUMN_LENGTH, "255")//$NON-NLS-1$
+                .type().stringType().noDefault().endRecord();
+        schemaReject.schema.setValue(s);
+    }
+
```
francisu: No Talend6 stuff.

francisu:
If not all of these are available with SchemaConstants, please file a Daikon JIRA to request they be added. Please convert everything you have to use SchemaConstants. @RyanSkraba can you have a look?

jzhao-talend:
@francisu 
I have create a jira  https://jira.talendforge.org/browse/TDKN-39

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforcebulkexec/TSalesforceBulkExecProperties.java
RyanSkraba: The TALEND6_COLUMN_TALEND_TYPE is unnecessary and will be picked up by the Avro type (stringType). 

Do we care what the TALEND6_COLUMN_LENGTH is?

TALEND6_IS_READ_ONLY is a bit more complicated -- this might be moved to SchemaConstants to indicate that a Schema or field is "locked", which should only be the case when a component is deciding on an outgoing schema.

Can we consider that if a Schema is "locked" that all of its fields are locked as well?

I don't know what the custom column property does in the studio.

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforcegetservertimestamp/TSalesforceGetServerTimestampProperties.java
RyanSkraba: The TALEND6_COLUMN_TALEND_TYPE is unnecessary and will be picked up by the Avro type (the logical type shown in the deleted code).

Do we care about TALEND6_COLUMN_LENGTH?  How about if we always set it to the length of the pattern in MetadataToolAvroHelper?

Is it mandatory that this schema is read-only?  If the user changes the column name to "MyTimestamp" or to type Long, do we care?  The component should be able to handle these situations.

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforceoutput/TSalesforceOutputProperties.java
RyanSkraba: The TALEND6_COLUMN_TALEND_TYPE is unnecessary and will be picked up by the Avro type.  But did you mean for the errorCode to be defined as an Avro intType yet a "id_String"?

I see here that if a Schema is "locked" that the fields can still be modified.  Does this mean that a "locked" field can't have its name or type changed, but a "locked" schema can't add or remove fields?

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforcebulkexec/TSalesforceBulkExecProperties.java
jzhao-talend: >>TALEND6_COLUMN_TALEND_TYPE 
ok. i would remove this one.

>> TALEND6_COLUMN_LENGTH 
This should not be used in salesforce, but when link a output line with a db output (like tMysqlOutput) . It would be used to created table....
And this should be replaced by SchemaConstants.TALEND_COLUMN_DB_LENGTH

>>TALEND6_IS_READ_ONLY 
This is control the table and field can be modified or not

>> Can we consider that if a Schema is "locked" that all of its fields are locked as well?
But I am not sure about this one. Need help from @nrousseau 

>>TALEND6_COLUMN_TALEND_TYPE
I can only sure the column color is different. Not sure the  other usage

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforcegetservertimestamp/TSalesforceGetServerTimestampProperties.java
jzhao-talend: **Is it mandatory that this schema is read-only?**
Yes. I just keep the old component behavior. Seems sure we can make open.

PR #64, opened 2016-04-29
TDI-35766: Added integration tests, removed not needed test class, fi…

chmyga:
…xed writer datacount result.

Added integration tests, but added Ignore annotation for them, as we currently don't have any test server.
Fixed TSplunkEventCollectorWriter returning result, found during integration tests.


components-splunk/src/test/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriterTestIT.java
```diff
@@ -0,0 +1,196 @@
...
+
+    private FieldAssembler<Schema> addField(FieldAssembler<Schema> record, String name, Class<?> type, AvroRegistry avroReg) {
+        Schema base = avroReg.getConverter(type).getSchema();
+        FieldBuilder<Schema> fieldBuilder = record.name(name);
+        fieldBuilder.type(AvroUtils.wrapAsNullable(base)).noDefault();
+        return record;
+    }
+
+    private static final class TestingRuntimeContainer implements RuntimeContainer {
+
```
francisu: There is a DefaultRuntimeContainerImpl class you can extend that does this. It's for testing.

components-splunk/src/test/java/org/talend/components/splunk/TSplunkEventCollectorTestIT.java
```diff
@@ -89,4 +121,9 @@ public void testAllRuntimes() {
         ComponentTestUtils.testAllRuntimeAvaialble(getComponentService());
     }
 
+    protected ComponentProperties checkAndAfter(Form form, String propName, ComponentProperties props) throws Throwable {
+        assertTrue(form.getWidget(propName).isCallAfter());
+        return getComponentService().afterProperty(propName, props);
+    }
+
```
francisu: This should be moved to AbstractComponentTest. Not sure why we are not using that (maybe you were told not to?) but we do want to be able to have a common superclass for the tests. This might be something for @sgandon  and I to discuss.

components-splunk/src/test/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriterTestIT.java
francisu: Please just fix the above thing and then I will merge it. Don't worry about my previous comments, we need to do something to improve the class hierarchy for the tests so they correctly run in both Spring and OSGi without the component developers having to do something.

PR #68, opened 2016-05-03
Bugfix/tup 4491 add input connector for salesforce output

nrousseau:


sgandon:
the Components project does not compile, Splunk is not up to date. I'll commit a fix for this before merging it.

nrousseau:
strange.. means in fact master didn't work before even, since of course i did not modify this one at all

PR #69, opened 2016-05-03
TDI-35441: tDatasetInput / tDatasetOutput Components

stavytskyi:


components-dataprep/pom.xml
```diff
@@ -0,0 +1,132 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.talend.components</groupId>
+        <artifactId>components-parent</artifactId>
+        <version>0.7.0-SNAPSHOT</version>
```
pteyssier: The version should be 0.8.0

components-dataprep/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+    </parent>
+
+    <name>Components - Data Preparation</name>
+    <groupId>org.talend.components</groupId>
+    <artifactId>components-dataprep</artifactId>
+    <version>0.7.0-SNAPSHOT</version>
+    <properties>
+        <components.api.version>0.7.0-SNAPSHOT</components.api.version>
```
sgandon: here too, version 0.8.0-SNAPSHOT

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepAvroRegistry.java
```diff
@@ -0,0 +1,261 @@
...
+        for (DataPrepField field : in) {
+
+            Schema.Field avroField = new Schema.Field(field.getColumnName(), inferSchema(field), null, field.getContent());
+
+            switch (field.getType()) {
+                case "date":
+                    avroField.addProp(SchemaConstants.TALEND_COLUMN_PATTERN, "yyyy-MM-dd");
+                    break;
+//                case datetime:
+//                    avroField.addProp(SchemaConstants.TALEND_COLUMN_PATTERN, "yyyy-MM-dd'T'HH:mm:ss'.000Z'");
```
sgandon: please remove the commented lines.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepAvroRegistry.java
```diff
@@ -0,0 +1,261 @@
...
+    private Schema inferSchemaField(DataPrepField field) {
+        Schema base;
+        switch (field.getType()) {
+            case "boolean":
+                base = AvroTypes._boolean();
+                break;
+            case "double":
+                base = AvroTypes._double();
+                break;
+//            case "numeric":
```
sgandon: please remove the commented lines.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepAvroRegistry.java
```diff
@@ -0,0 +1,261 @@
...
+                base = AvroTypes._string();
+                break;
+        }
+//        base = field.getNillable() ? AvroUtils.wrapAsNullable(base) : base;
+
+        return base;
+    }
+
+    /**
+     * A helper method to convert the String representation of a datum in the Salesforce system to the Avro type that
```
sgandon: javadoc should not refer to Salesforce, a bad cut/paste I guess  ;) .

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepAvroRegistry.java
```diff
@@ -0,0 +1,261 @@
...
+            return new StringToDateConverter(f);
+        } else if (AvroTypes.isSameType(fieldSchema, AvroTypes._string())) {
+            return super.getConverter(String.class);
+        }
+        throw new UnsupportedOperationException("The type " + fieldSchema.getType() + " is not supported."); //$NON-NLS-1$ //$NON-NLS-2$
+    }
+
+    // TODO(rskraba): These are probably useful utility items.
+
+    public static abstract class AsStringConverter<T> implements AvroConverter<String, T> {
```
sgandon: The following methods have nothing specific to Datapred, they should be located in daikon for everyone to use.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepAvroRegistry.java
stavytskyi: Yes, it should be refactored too because difference only in one method.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepAvroRegistry.java
stavytskyi: I've left this just like notice, can I add todo??

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepAvroRegistry.java
stavytskyi: Yes, I think all class should be refactored, a lot of method is reused I think It can be in daikon.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
```diff
@@ -0,0 +1,154 @@
...
+        this.dataSetName = dataSetName;
+    }
+
+    HttpResponse connect() {
+        Request request = Request.Post(url+"/login?username="+login+"&password="+pass);
+        HttpResponse response = null;
+        try {
+            response = request.execute().returnResponse();
+        } catch (IOException e) {
+            e.printStackTrace();
```
sgandon: Never use *e.printStackTrace*, please handle the exceptions properly by at least login them using slf4j if you really want to ignore them.

sgandon: This is a low level class, so I would suggest you rethrow those exceptions to handle them in a higher level

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+        authorisationHeader = response.getFirstHeader("Authorization");
+        return response;
```
sgandon: response may be null here, is it excepted, please add some javadoc.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
```diff
@@ -0,0 +1,154 @@
...
+        ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.configure(DeserializationConfig.Feature.FAIL_ON_UNKNOWN_PROPERTIES,
+                false);
+        SourceDataSet dataSet = null;
+        try {
+            dataSet = objectMapper.readValue(current.getEntity().getContent(), SourceDataSet.class);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return dataSet.getRecords();
```
sgandon: NPE here is exception is caught.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
```diff
@@ -0,0 +1,154 @@
...
+            logout();
+        }
+//        System.out.println(test);
+        MetaData metaData = null;
+        try {
+            metaData = objectMapper.readValue(test, MetaData.class);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return metaData.getColumns();
```
sgandon: NPE here if exception is caught.

stavytskyi:
I think we cannot handle limits in tDataSetOutput components, because we cannot handle how many times called write method in TDataSetOutputWriter. Limits should handle how many rows(IndexRecord) will passed to DataPrep.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
sgandon: I think François advised you to use a Dataset row iterator here because we don't want to load all the data into the memory and we want to handle them as a stream flow.
This will not be robust enough for big datasets.

components-dataprep/src/main/java/org/talend/components/dataprep/TDataSetInputSource.java
```diff
@@ -0,0 +1,101 @@
...
+        if (connectionHandler == null) {
+            connectionHandler = new DataPrepConnectionHandler(properties.url.getStringValue(), properties.login.getStringValue(),
+                    properties.pass.getStringValue(), "read", properties.dataSetName.getStringValue());
+            return connectionHandler;
+        } else
+            return this.connectionHandler;
+    }
+
+    public ValidationResult validate(RuntimeContainer container) {
+//        if (getConnectionHandler().validate())
```
sgandon: I think you should be at least checking the connection here to make sure dataprep server is reachable and the credentials are ok.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
stavytskyi: Yes, It was my thoughts too, but I didn't have dataprep lib in repo on Friday. so If I'd used locally jar, build would fail.

components-dataprep/src/test/java/org/talend/components/dataprep/TDataPrepConnectionHandlerTest.java
```diff
@@ -0,0 +1,80 @@
...
+        assertTrue(connectionHandler.validate());
+    }
+
+    @Test
+    public void readSchema() throws IOException {
+        DataPrepConnectionHandler connectionHandler = new DataPrepConnectionHandler(
+                URL, LOGIN, PASS, "read", "0d3df0df4a4aca0529ef5755bd03519adb115248");
+        System.out.println(connectionHandler.connect());
+        for (Column column: connectionHandler.readSourceSchema()) {
+            System.out.println(column);
```
sgandon: what are you testing here, all test case should have asserts or expect an exception to be thrown.
same comment for the tests below.

stavytskyi: Yes, It is not test, this code I've used for handling day to day dataprep changes. I've added it to ignore. Should I remake now?

components-dataprep/src/main/java/org/talend/components/dataprep/TDataSetOutputWriter.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Override
+    public void open(String uId) throws IOException {
+        this.uId = uId;
+        connectionHandler.connect();
+    }
+
+    @Override
+    public void write(Object datum) {
+        if (datum == null) {
```
sgandon: I don't understand this at all and there is not test case for this.
the first record to be stored will have a string of field names  +  a string values separated with a comas.
then the second record will only have a string a values, 
and you will accumulate them in the memory and really send them to data prep is the close, if I am right.

This is not right,  every record shall be sent one by one in a stream, otherwise there is not way you can handle 1 milllion lines and 100 columns.

components-dataprep/src/main/java/org/talend/components/dataprep/TDataSetOutputWriter.java
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+        data.append(row);
+        data.append("\n");
+        counter++;
+    }
+
+    @Override
+    public WriterResult close() {
+        System.out.println(data.toString());
```
sgandon: no *System.out.println* is allowed here, use log4j.

components-dataprep/src/main/java/org/talend/components/dataprep/TDataSetOutputWriter.java
stavytskyi: Yes, you are almost right. Every record and column names that I take from schema should be sent one by one in a stream.

sgandon:
Of course you can handle limit, just count the number of time the Writer.write() is called and you then do not sent the rest of the data for subsequent calls.

sgandon:
To sum up,
* use streaming on the input and output otherwise the goal of handling 1 000 000 lines with 100 columns will never work.
* never use *e.PrintStackTrace*, you must handle the exception or throw them, usually you will want to throw them back to the system.
* never use *System.out.println*, use *slf4j* log APIs.
* core classes such as input read and output writer need tests.
* but not test that do not assert anything.
* and remove commented code please.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepAvroRegistry.java
sgandon: Yes //TODO is fine

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
sgandon: If you have any issue with a dataprep lib please let them know and they will help you. they are not aware of any issue on your side.

stavytskyi: François said me on Friday that he would publish library in open source repo. I have just reminded him.

stavytskyi:
Task for 6.2.1:
- rewrite DataPrepConnectionHandler for using apache cxf (also add streaming for output component).
- refactor DataPrepAvroRegister (reflect daikon and salesforce component).
- add streaming for json like it works in dataprep DataSetRowIterator.
- add right handling for 'numeric', 'any', 'date' dataprep type.
- write tests with docker.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
pteyssier: You should use logger here.

PR #70, opened 2016-05-03
Tdkn 35 make property generic

sgandon:
the Property was made generic : https://github.com/Talend/daikon/pull/32
This PR is to accommodate this change.
Please review.


components-api/src/main/java/org/talend/components/api/properties/ComponentPropertyFactory.java
```diff
@@ -43,15 +43,15 @@ public static Property newReturnsProperty() {
      * @param name the name of the returns property
      * @return a {@link Property}
      */
-    public static Property newReturnProperty(Property returns, Property.Type type, String name) {
+    public static <T> Property<T> newReturnProperty(Property<String> returns, Class<T> type, String name) {
```
gonchar-ivan: I think "returns" parameter name is unclear. It is unclear that it is parent Property for all returns properties.
Maybe "parentReturns" will be better. Or simply "parent".

sgandon: Thank you for the comment, you are right, I have done a lot of modifications since this PR and I have changed the factory method for newReturnProperty.
to this :
public static <T> Property<T> newReturnProperty(Property<String> returns, Property<T> returnProp)
What do you think ?

gonchar-ivan: Yes. I think it's good. Thank you.

PR #74, opened 2016-05-04
TUP-4428: Handle reject record in component side.

wwang-talend:
https://jira.talendforge.org/browse/TUP-4428


francisu:
DO you need this for 6.2.0? If so, I will merge it, but I'm not sure about the DataRejectException, and there are no Junit tests for this. If not, I need to think about the exception with Seb and there need to be some tests.

PR #75, opened 2016-05-04
Feature/tdi 35765 jira components (WIP)

gonchar-ivan:


components-jira/pom.xml
```diff
@@ -0,0 +1,180 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.talend.components</groupId>
+		<artifactId>components-parent</artifactId>
+		<version>0.7.0-SNAPSHOT</version>
```
sgandon: use version 0.8.0-SNAPSHOT

components-jira/pom.xml
```diff
@@ -0,0 +1,180 @@
...
+	</parent>
+
+	<properties>
+		<components.api.version>0.7.0-SNAPSHOT</components.api.version>
```
sgandon: use version 0.8.0-SNAPSHOT

components-jira/src/main/java/org/talend/components/jira/connection/Rest.java
```diff
@@ -0,0 +1,196 @@
...
+    public String post(String resource, String body) throws ClientProtocolException, IOException {
+        Request post = Request.Post(hostPort + resource).bodyString(body, contentType);
+        for (Header header : headers) {
+            post.addHeader(header);
+        }
+        Content content = post.execute().returnContent();
+        if (content != null) {
+            return content.asString();
+        }
+        return "204 no content";
```
sgandon: This is a strange return, why don't you return null ?

gonchar-ivan:
Fixed according Sebastian's comments

PR #76, opened 2016-05-04
Feature/tup 4442 update table

nrousseau:


sgandon:
I think you could add some test cases for the *getHeader()* and *getValue()* methods.

pteyssier:
We will create tests in TUP-4480

PR #81, opened 2016-05-05
TCOMP-154: Add return the return value for salesforce input/output components

jzhao-talend:
https://jira.talendforge.org/browse/TCOMP-154
Add return value for salesforce


jzhao-talend:
The second commit with wrong comment

PR #88, opened 2016-05-09
TUP-4541: fix problem of refresh, sub-forms are not refreshed when call

nrousseau:


nrousseau:
Commit integrated without recheck first with seb since we didn't get time.
To recheck after again

PR #99, opened 2016-05-13
TUP-4544: Add support foreign key in SOQL for tSalesforceInput

jzhao-talend:
https://jira.talendforge.org/browse/TUP-4544

### DIFF
--- a/components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SObjectAdapterFactory.java
+++ b/components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SObjectAdapterFactory.java
@@ -1,5 +1,6 @@
 package org.talend.components.salesforce.runtime;
 
+import com.sforce.ws.bind.XmlObject;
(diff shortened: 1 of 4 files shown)


components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SObjectAdapterFactory.java
```diff
@@ -48,10 +56,17 @@ public IndexedRecord convertToAvro(SObject value) {
...
+        private Map<String, Object> valueMap;
 
         public SObjectIndexedRecord(SObject value) {
-            this.value = value;
+            Iterator<XmlObject> fields = value.getChildren();
+            while (fields.hasNext()) {
+                try {
+                    processXmlObject(fields.next(), null, null);
+                } catch (IOException e) {
+                    e.printStackTrace();
```
sgandon: please make sure you handle exception properly.

PR #100, opened 2016-05-13
TDI-36100: Write data to dataprep server with stream

stavytskyi:


pteyssier:
Need to rebase on latest commits from Vincent Jaouen

PR #102, opened 2016-05-13
Feature/tdi 35765 jira components

gonchar-ivan:
Finished initial Jira Input component
https://jira.talendforge.org/browse/TDI-35890

Unit-tests and Integration tests are not finished, because I need to switch on Jira Output component implementation


components-jira/pom.xml
```diff
@@ -92,7 +92,18 @@
...
 		</dependency>
-
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<version>1.10.19</version>
+			<scope>test</scope>
+			<exclusions>
+				<exclusion>
+					<groupId>org.hamcrest</groupId>
```
gonchar-ivan: Mockito depends on old hamcrest version.
Used version from parent pom.
Maybe hamcrest version should be added to <dependencyManagement> of parent pom

components-jira/src/main/java/org/talend/components/jira/runtime/JiraSource.java
```diff
@@ -33,28 +32,73 @@
...
+     * Saves component properties in this object
      * 
      * @param container runtime container
-     * @param properties
+     * @param properties component properties
      */
     @Override
     public void initialize(RuntimeContainer container, ComponentProperties properties) {
-        // FIXME could it throw cast exception?
-        this.properties = (TJiraInputProperties) properties;
```
gonchar-ivan: As I understand it is bad to store properties instance in Source or Sink. (Because of Serialization)
Am I right?

components-jira/src/test/java/org/talend/components/jira/tjirainput/TJiraInputPropertiesTest.java
```diff
@@ -0,0 +1,338 @@
...
+
+    /**
+     * Sets up mocks required for tests
+     */
+    @Before
+    public void setUpMocks() {
+        userPasswordWidget = mock(Widget.class);
+        jqlWidget = mock(Widget.class);
+        projectIdWidget = mock(Widget.class);
+
```
francisu: I don't understand why you are using mocks for Widget, Form, and properties. Why not use the real objects?

components-jira/src/main/java/org/talend/components/jira/runtime/JiraSource.java
francisu: I would not worry about the classcastexception, and it's no problem to store the properties in Source or Sink as they can be serialized.

components-jira/pom.xml
francisu: I think this is a good idea. Can you do that?

gonchar-ivan: Yes. I will do it in separate pull request

components-jira/src/test/java/org/talend/components/jira/tjirainput/TJiraInputPropertiesTest.java
gonchar-ivan: Reasons:

1) Make unit-tests independent
When I implement Unit-tests I try to follow requirement:
Unit-tests should be independent. If something was changed in dependency class (Widget, Form, etc), it should not affect test result.

This could be achieved by using mocks. I think changes in Widget, Form classes should be covered (tested) by their own tests.
Let's assume something was broken in Widget class and we got unit-tests fails.
In case of using mocks: only Widget unit-tests will fail and we can easily identify that the problem is with Widget class
In case of using real objects: all unit-tests, where real Widget is used will fail. We should spent some time to identify that the problem is with Widget class.
So, advantage is - easy to identify where is the problem.
Disadvantage is - using real objects adds additional coverage to dependency classes, but using mock doesn't add it.
(If Widget unit-tests doesn't cover some case, it could be covered by other classes tests, which are used real Widget class)

2) Sometimes real objects requires complicated configuration (using its API) to make its method return needed value.
I need to add additional method for my test class... something like prepateTestParameter, before I can use it.
But it is easy to make mock to return needed value.

I agree, in case of Widget, Properties and Forms it is not difficult to configure them. But for other classes, it may be difficult.
So, here I've used mocks from force of habit

sgandon: I am fine with this.

francisu: I'm not. If the underlying API you depends on changes in a way that breaks compatibility, then we want to know it and the code may have to be fixed. Mocks will hide that. And further, they are using the classes in ways not intended, which may require maintenance even if the classes are changed in a compatible way. Please remove the mocks and configure them as normal. I think adding the mocks is just additional complexity. We have done a lot of work to make sure these classes are easy to configure and use.

There are cases where mocks might be applicable, but I don't think this is one of them.

gonchar-ivan: @francisu 
It seems, I understand your point of view.
When framework is changed, it could require changes in component.
In case of using real objects, we get failed tests. So we are forced to make changes in component.
In case of using mocks, we get passed tests. So we don't know about errors until deploy in Platform.

I didn't consider this case fully. My bad. 
I just thought about case, if somebody makes a mistake in framework classes, then mistakes should be covered by framework unit-tests and shouldn't affect the result of component unit-tests. But I didn't consider case, when we need to change framework logic.

Other my idea was to use mocks in unit-tests and use real objects in integration tests to check the same thing. But in this case I need to duplicate tests. It's not a good idea.

So, I agree with you. I will remove mocks.

gonchar-ivan: Removed mocks

PR #105, opened 2016-05-16
TDI-36100: Write data to dataprep server with stream

stavytskyi:
I've rebased and made changes for live dataset. I think @sgandon should review too. I'm going to rewrite all test. It's almost not relevant.


components-dataprep/src/main/java/org/talend/components/dataprep/TDataSetOutputWriter.java
```diff
@@ -22,16 +22,16 @@
 import org.talend.daikon.avro.AvroRegistry;
 import org.talend.daikon.avro.IndexedRecordAdapterFactory;
 
-import java.io.IOException;
+import java.io.*;
```
pteyssier: Avoid multiple import. import only what is necessary. Is IDEA doing that?

components-dataprep/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.talend.components</groupId>
         <artifactId>components-parent</artifactId>
-        <version>0.9.0-SNAPSHOT</version>
+        <version>0.8.0-SNAPSHOT</version>
```
pteyssier: a PR should not change the version in POM

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
```diff
@@ -89,26 +100,40 @@ DataPrepStreamMapper readDataSetIterator() throws IOException {
...
-    void createInLiveDataSetMode(String data) throws IOException {
+    OutputStream createInLiveDataSetMode() throws IOException {
 
         LOGGER.debug("DataSet name: " + dataSetName);
 
         // in live dataset, the request is a simple post to the livedataset url
-        Request request = Request.Post(url);
-        request.bodyString(data, ContentType.create(ContentType.TEXT_PLAIN.getMimeType(), StandardCharsets.UTF_8));
-        HttpResponse response = request.execute().returnResponse();
-        LOGGER.debug("Create request response: {}", response);
```
pteyssier: Not sure if this logger should have been removed

components-dataprep/src/test/java/org/talend/components/dataprep/TDataPrepConnectionHandlerTest.java
```diff
@@ -158,9 +142,8 @@ public void sendDataWithStream() throws IOException {
     }
 
     @Test
-    public void sendDataWithStreamApache() throws IOException{
-//        WebClient.create(URL).path(path).accept(MediaType.APPLICATION_XML)
-//                .query(typeParam, type).query("_type", "xml").get(Products.class);
+    public void some() {
+        InputStream inputStream;
```
pteyssier: ?

components-dataprep/src/test/java/org/talend/components/dataprep/TDataPrepConnectionHandlerTest.java
```diff
@@ -144,10 +128,10 @@ public void sendDataWithStream() throws IOException {
...
 
         String body = "col1, col2, col3\ntest1, test2, test3\ntest4, test5, test6";
-        DataOutputStream outputStream = new DataOutputStream(connection.getOutputStream());
+        OutputStream outputStream = new DataOutputStream(connection.getOutputStream());
 
         for (int i = 0; i < 100000; i++) {
-            outputStream.writeBytes(body);
+            outputStream.write(body.getBytes());
             outputStream.flush();
             System.out.println(i);
```
sgandon: avoir using system.out.println even in test cases, use a slf4j logger instead.

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
```diff
@@ -80,7 +90,8 @@ boolean validate() throws IOException {
     }
 
     DataPrepStreamMapper readDataSetIterator() throws IOException {
-        Request request = Request.Get(url + "/api/datasets/" + dataSetName + "?metadata=false").addHeader(authorisationHeader);
+        Request request = Request.Get(url+ "/api/datasets/" + dataSetName + "?metadata=false").
+                addHeader(authorisationHeader);
         HttpResponse current = request.execute().returnResponse();
         DataPrepStreamMapper dataPrepStreamMapper = new DataPrepStreamMapper(current.getEntity().getContent());
```
sgandon: You should be handling the response here cause if the get fails you'll have and exception when calling the getEntity().getContent()
Please after each request you need to test the response result and handle it accordingly.
May be you could throw an IOException when the request fails.
The same comment goes for every method that does a remote request.

francisu:
I'm OK with this, but it looks like there is one issue for Pierre not resolved.

pteyssier:
ok on my side

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
```diff
@@ -56,20 +56,31 @@
     HttpResponse connect() throws IOException {
         Request request = Request.Post(url + "/login?username=" + login + "&password=" + pass);
         HttpResponse response = request.execute().returnResponse();
-        LOGGER.debug("Connect Response: " + response.toString());
         authorisationHeader = response.getFirstHeader("Authorization");
+        if (returnStatusCode(response) != HttpServletResponse.SC_OK && authorisationHeader != null) {
+            LOGGER.error("Failed to login to Dataprep serrver: {}", returnStatusCode(response));
+            // TODO i18n
+            throw new IOException("Failed to login to Dataprep serrver: " + returnStatusCode(response));
```
sgandon: serrver is only having one 'r' ;)

stavytskyi: Yes, thank you :)

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
```diff
@@ -72,43 +93,59 @@ boolean validate() throws IOException {
         int statusLogout;
         statusLogin = returnStatusCode(connect());
         statusLogout = returnStatusCode(logout());
-        if (statusLogin == STATUS_OK && statusLogout == STATUS_OK) {
-            return true;
-        } else {
-            return false;
-        }
+        return statusLogin == STATUS_OK && statusLogout == STATUS_OK;
```
sgandon: if there is an error here you should log it before returning the validation status.

stavytskyi: ok, so I will remove exception handling from validate() in Source or Sink and add it here. Is it ok?

components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
```diff
@@ -51,64 +56,106 @@
...
+        }
         return response;
     }
 
     private int returnStatusCode(HttpResponse response) {
         return response.getStatusLine().getStatusCode();
     }
 
-    boolean validate() throws IOException {
+    boolean validate() {
```
sgandon: I think this method should not return a boolean and throw an exception with the proper message if there is an validation issue so that the called of this method knows why the validation failed.
For example in the SourceValidate you can return the ValidationResult with the proper message so that the user knows about what went wrong.

components-dataprep/src/main/java/org/talend/components/dataprep/TDataSetInputSource.java
```diff
@@ -68,10 +68,13 @@ private DataPrepConnectionHandler getConnectionHandler() {
...
     public ValidationResult validate(RuntimeContainer container) {
-        if (getConnectionHandler().validate())
-            return ValidationResult.OK;
-        else
-            return new ValidationResult().setStatus(ValidationResult.Result.ERROR);
+        try {
+            getConnectionHandler().validate();
+        } catch (IOException e) {
+            return new ValidationResult().setStatus(ValidationResult.Result.ERROR)
+                    .setMessage(e.getMessage());
```
sgandon: yes much better ;)

sgandon:
Thanks for your work.

PR #108, opened 2016-05-17
TDI-35766: added docker plugin for integration tests.

chmyga:
Added docker maven plugin to run container with splunk image. It will be started on pre-integration-test phase and stopped and removed on post-integration-test phase.
Defined a system property in pom.xml to be used in integration tests. Removed Ignore annotation from integration tests to run them during build.

### DIFF
--- a/components-splunk/pom.xml
+++ b/components-splunk/pom.xml
@@ -15,6 +15,8 @@
     <version>0.2.0-SNAPSHOT</version>
     <properties>
         <components.api.version>0.9.0-SNAPSHOT</components.api.version>
+        <splunk.webui.port>8000</splunk.webui.port>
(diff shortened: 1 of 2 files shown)


chmyga:
DOCKER_HOST and DOCKER_CERT_PATH should be set in environment, where maven build is run. DOCKER_HOST - url of docker host(if it's remote machine, for example VM on windows machine or remote network machine). it can be not set if the OS is Linux and docker is running on the local machine.
DOCKER_CERT_PATH - path to docker certificates.

francisu:
This needs to be set up in the build environment before we can merge this. And there should be something in a README in your module that talks about the required setup.

I will ask stefan baryakov to have a look at this.

PR #109, opened 2016-05-17
Bugfix/tcomp 195 bundle plugin version

gonchar-ivan:
Fixed bug with maven bundle plugin version
https://jira.talendforge.org/browse/TCOMP-195

Doesn't remove plugin version in dataprep component, because it fails build.
Dataprep is not using relative path to parent pom.xml


components-api/pom.xml
```diff
@@ -202,8 +202,6 @@
 			<plugin>
 				<groupId>org.apache.felix</groupId>
 				<artifactId>maven-bundle-plugin</artifactId>
-				<version>3.0.0</version>
-				<extensions>true</extensions>
```
gonchar-ivan: I'm not sure it is correct to move <extensions> tag to parent pom.xml. But it seems that it works

components-parent/pom.xml
```diff
@@ -216,7 +222,7 @@
 						<configuration>
 							<tasks>
 								<echo file="target/classes/META-INF/maven/dependencies.txt"
-									message="${groupId}:${artifactId}:jar:${version}:runtime"
```
gonchar-ivan: It fixes warnings during maven build.
dependencies.txt is not changed

components-parent/pom.xml
```diff
@@ -185,6 +185,12 @@
                         <tag>${project.artifactId}-${project.version}</tag>
                     </configuration>
                 </plugin>
+                <plugin>
+					<groupId>org.apache.felix</groupId>
+					<artifactId>maven-bundle-plugin</artifactId>
+					<version>3.0.1</version>
```
gonchar-ivan: What version will maven use if we don't specify version here?
The latest one?
Do we need to specify fixed version here?

francisu:
The archetype test fails when I try this. Have you checked that? mvn clean install does not work.

gonchar-ivan:
I've checked that (components-api/ mvn clean install). Build failed. But it failed on master too for me. So I thought that it is not because of my changes.

After your I've investigated a problem more deeply. I had failed build on master because of error in Maven 3.3.9
https://issues.apache.org/jira/browse/ARCHETYPE-488

I've fixed by following workaround:
copy apache-maven-3.3.9\bin\mvn.cmd -> mvn.bat
This has fixed my build on master and showed a real problem with my change.
I moved <extensions>true</extensions> parameter from bundle plugin configuration to parent pom.xml. 
But archetype pom.xml has components-parent version 0.8.3 and has no relative path to parent pom.xml (to which I moved this parameter). This was a cause.

So I need to return it back

components-api/pom.xml
gonchar-ivan: It caused a problem to archetype project. So I returned this tag back to component projects

components-jira/pom.xml
```diff
@@ -117,8 +117,6 @@
 			<plugin>
 				<groupId>org.apache.felix</groupId>
 				<artifactId>maven-bundle-plugin</artifactId>
-				<version>2.5.4</version>
-				<extensions>true</extensions>
```
sgandon: Why  then don't you keep the <extensions> element in this pom but kept it in others?

gonchar-ivan:
Also I can't build components-salesforce and components-osgitest on my side.
(I think because I can't connect to real Salesforce)
Are there some tips how to fix it?

components-parent/pom.xml
sgandon: yes this is perfect, we need to have the version here so that every pom that use this as a parent will use the same version.  This is much better for consistency.

components-jira/pom.xml
gonchar-ivan: hm... because it's my component and I know about the problem. (parent pom also has extensions = true element).
If I need to change version of parent pom, then I know I need to add extensions element.
Maybe I'am wrong.

gonchar-ivan: Fixed. Removed extensions element from components, which are using 0.10.0 version of components-parent

PR #112, opened 2016-05-18
TDI-35874: Create unit tests for tDataSetInput/tDataSetOutput component

stavytskyi:


components-dataprep/src/main/java/org/talend/components/dataprep/TDataSetInputProperties.java
```diff
@@ -99,6 +97,7 @@ public void setupLayout() {
     }
 
     private String removeQuotes(String str) {
```
sgandon: why would you have to remove any quotes ?. The Studio should take care of it. The component should never have the knowledge of what the client does.

components-dataprep/src/test/java/org/talend/components/dataprep/DataPrepStreamMapperTest.java
```diff
@@ -13,7 +15,7 @@
     //datasetid=7fa267c6-87c8-4ba2-9b24-fe300bf2f163
 
     @Test
-    public void metaDataMappingTest() throws IOException {
+    public void testMetaDataMapping() throws IOException {
         InputStream inputStream = DataPrepStreamMapperTest.class.getResourceAsStream("metadata.json");
         DataPrepStreamMapper streamMapper = new DataPrepStreamMapper(inputStream);
         MetaData metaData = streamMapper.getMetaData();
```
sgandon: where is your test ?
you must assert things here, check that the Metadata object contains what you expect.

components-dataprep/src/test/java/org/talend/components/dataprep/DataPrepStreamMapperTest.java
```diff
@@ -36,4 +38,11 @@ public void rowDataSetIteratorTest() throws IOException {
             }
         streamMapper.close();
     }
```
sgandon: here too you need asserts.

components-dataprep/src/main/java/org/talend/components/dataprep/TDataSetInputProperties.java
stavytskyi: I think it related to design time invocation. Studio doesn't remove quotes before call afterFetchSchema() but remove in other situations.

components-dataprep/src/test/java/org/talend/components/dataprep/TDataSetOutputSinkTest.java
```diff
@@ -0,0 +1,72 @@
...
+        definition = (TDataSetOutputDefinition) (componentService.getComponentDefinition("tDatasetOutput"));
+        outputSink = (TDataSetOutputSink) definition.getRuntime();
+    }
+
+    @Test
+    public void testCreateWriteOperation() throws Exception {
+        TDataSetOutputProperties properties = (TDataSetOutputProperties) definition.createProperties();
+        properties.setValue("url", "http://127.0.0.1");
+        outputSink.initialize(null, properties);
+        outputSink.createWriteOperation();
```
sgandon: please not test without assert except if you excpect some exception to be thrown.

components-dataprep/src/main/java/org/talend/components/dataprep/TDataSetInputProperties.java
sgandon: then this is a Studio bug and you must report it.

stavytskyi: I will make bug report about that. Should I live it befere bug will be fixed?

sgandon: yes, please enter a jira for you to think about removing it later.

sgandon:
I'll merge this once the build passes, we have issues on our ci.

PR #117, opened 2016-05-20
TDP-1879: dataprep component now adds its client ID for dataprep logi…

vjaouen:
…n & logout calls

### DIFF
--- a/components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
+++ b/components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
@@ -54,7 +54,7 @@ public class DataPrepConnectionHandler {
     }
 
     HttpResponse connect() throws IOException {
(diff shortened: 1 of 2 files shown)


components-dataprep/src/main/java/org/talend/components/dataprep/DataPrepConnectionHandler.java
```diff
@@ -74,7 +74,7 @@ HttpResponse logout() throws IOException{
                 urlConnection.disconnect();
             }
         } finally {
-            Request request = Request.Post(url + "/logout").addHeader(authorisationHeader);
+            Request request = Request.Post(url + "/logout?client-app=STUDIO").addHeader(authorisationHeader);
```
sgandon: Is it expected that the STUDIO value is in upper-case for logout and lower case for login ?

vjaouen: Case is ignored in dataprep

PR #118, opened 2016-05-20
TDI-36288: Clean code

stavytskyi:


stavytskyi:
Please, don't merge the pull request. I would like also separate classes for packages.

components-dataprep/src/main/java/org/talend/components/dataprep/runtime/RuntimeProperties.java
```diff
@@ -0,0 +1,87 @@
...
+// ============================================================================
+package org.talend.components.dataprep.runtime;
+
+import java.io.Serializable;
+
+public class RuntimeProperties implements Serializable {
+
+    private static final long serialVersionUID = 3992422924211437022L;
+    private String ulr;
+    private String login;
```
francisu: Do you mean URL? Is not, what's a ulr?

components-dataprep/src/main/java/org/talend/components/dataprep/runtime/TDataSetInputSource.java
```diff
@@ -40,62 +45,70 @@
 public class TDataSetInputSource implements BoundedSource {
```
gonchar-ivan: I think, long class names are not good.
You may name this class DataSetSource. 
"Source" already means that it is "Input"
Also don't understand why we need "T" prefix

components-dataprep/src/main/java/org/talend/components/dataprep/runtime/TDataSetOutputSink.java
```diff
@@ -26,34 +30,34 @@
 
 public class TDataSetOutputSink implements Sink {
```
gonchar-ivan: You may name this class DataSetSink

components-dataprep/src/main/java/org/talend/components/dataprep/runtime/TDataSetOutputWriter.java
```diff
@@ -36,16 +38,17 @@
     private boolean firstRow = true;
     private WriteOperation<WriterResult> writeOperation;
     private int limit;
-    private String mode;
+    private DataPrepOutputModes mode;
 
-    TDataSetOutputWriter(WriteOperation<WriterResult> writeOperation, //
```
gonchar-ivan: You may name this class DataSetWriter

components-dataprep/src/main/java/org/talend/components/dataprep/runtime/RuntimeProperties.java
```diff
@@ -0,0 +1,87 @@
...
+// You should have received a copy of the agreement
+// along with this program; if not, write to Talend SA
+// 9 rue Pages 92150 Suresnes, France
+//
+// ============================================================================
+package org.talend.components.dataprep.runtime;
+
+import java.io.Serializable;
+
+public class RuntimeProperties implements Serializable {
```
gonchar-ivan: Properties suffix seems to be confusing for this class, because it doesn't extend Properties common class.

components-dataprep/src/main/java/org/talend/components/dataprep/runtime/TDataSetInputSource.java
gonchar-ivan: You may add DataSetSourceOrSink parent class, which contain common functionality for Source and Sink:
validate()
runtimeProperties
getSchema()
etc

stavytskyi: I know about that, I would like to make that in last refactoring.

PR #123, opened 2016-05-27
TDI-36288: Clean code and refactoring

stavytskyi:


components-dataprep/src/main/java/org/talend/components/dataprep/tdatasetoutput/TDataSetOutputProperties.java
```diff
@@ -47,36 +49,27 @@
...
-    public Property<String> login = PropertyFactory.newString("login");
-
-    public Property<String> pass = PropertyFactory.newString("pass");
-
-    public Property<String> url = PropertyFactory.newString("url");
-
-    public Property<Mode> mode = PropertyFactory.newEnum("mode", Mode.class);
-
+    public Property login = PropertyFactory.newString("login");
+    public Property pass = PropertyFactory.newString("pass");
```
jzhao-talend: @stavytskyi 
Seems you didn't use the official code format.
https://github.com/Talend/tools/blob/master/tools-java-formatter/src/main/resources/talend_java_eclipse_formatter.xml

components-dataprep/src/main/java/org/talend/components/dataprep/runtime/RuntimeProperties.java
```diff
@@ -0,0 +1,83 @@
...
+// You should have received a copy of the agreement
+// along with this program; if not, write to Talend SA
+// 9 rue Pages 92150 Suresnes, France
+//
+// ============================================================================
+package org.talend.components.dataprep.runtime;
+
+import java.io.Serializable;
+
+public class RuntimeProperties implements Serializable {
```
jzhao-talend: @stavytskyi 
Is this Class really needed in runtime ?
Why we just use "TDataSetInputProperties / TDataSetOutputProperties" ?

stavytskyi: When we develop components we should think about components will running in distributed systems. We cannot use "TDataSetInputProperties / TDataSetOutputProperties" as instance member of 'SourceOrSink' because it's not serializable and contains a lot of not needed information.

components-dataprep/src/main/java/org/talend/components/dataprep/tdatasetoutput/TDataSetOutputProperties.java
```diff
@@ -47,36 +49,27 @@
...
+    public Property url = PropertyFactory.newString("url");
+    public Property mode = PropertyFactory.newEnum("mode", DataPrepOutputModes.class);
```
jzhao-talend: @stavytskyi 
I think that is better change to :
> public Property**\<DataPrepOutputModes\>** mode = PropertyFactory.newEnum("mode", DataPrepOutputModes.class);

Then no need to be cast when you get the value

> runtimeProperties.setMode((DataPrepOutputModes)mode.getValue());

stavytskyi: I agree, fixed in next commit.

PR #127, opened 2016-05-31
TDKN-53 : moved Property into property package

sgandon:
see PR : https://github.com/Talend/daikon/pull/38
Please do not merge until the Studio has been adapted, it breaks the APIs.


sgandon:
already merged

PR #131, opened 2016-06-02
Bugfix/tup 4671 proxy setting issue

jzhao-talend:


components-common/src/main/java/org/talend/components/common/ProxyProperties.java
```diff
@@ -57,12 +57,13 @@ public void setupLayout() {
...
         refreshLayout(getForm(Form.MAIN));
+        refreshLayout(getForm(Form.REFERENCE));
     }
 
     @Override
     public void refreshLayout(Form form) {
         super.refreshLayout(form);
-        if (form.getName().equals(Form.MAIN)) {
+        if (form != null && form.getName().equals(Form.MAIN)) {
             boolean isUseProxy = useProxy.getValue();
```
francisu: You don't need the checks for form != null; the form will always be set.

jzhao-talend: @francisu 
Do you mean only for "ProxyProperties" or all I have added in this commit? I add this, just because of I get NPE when refresh proxy UI.
To avoid the NPE, I flow this modification https://github.com/Talend/daikon/commit/5abfd289b30ca5f6b7de3f192b3e7f30588569ea

**And seems proxy refresh NPE problem have been fixed yesterday.**

sgandon: yes it was fixed.

components-common/src/test/java/org/talend/components/common/runtime/ProxyPropertiesRuntimeHelperTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+        ProxyPropertiesRuntimeHelper.setProxy(proxyProperties, ProxyProperties.ProxyType.SOCKS);
+
+        assertEquals("true", System.getProperty("socksProxySet"));
+        assertEquals("192.168.32.162", System.getProperty("socksProxyHost"));
+        assertEquals("1080", System.getProperty("socksProxyPort"));
+        assertEquals("talend", System.getProperty("java.net.socks.username"));
+        assertNull(System.getProperty("java.net.socks.password"));
+
+        // Revert the modification about the Socks proxy during the test
```
sgandon: please place the revert methods into a **finally** block.

components-common/src/main/java/org/talend/components/common/ProxyProperties.java
francisu: I don't think @sgandon fix is right. refreshLayout should never be called with a null form. If that's happening, it's a bug somewhere, and you should track down where that's happening. I think the fix that Seb did for that should be backed out and it should get the NPE. And yes, this applies to all of the checks for form != null.

jzhao-talend: @francisu 
ok. I would remove them. Thanks.

components-common/src/test/java/org/talend/components/common/runtime/ProxyPropertiesRuntimeHelperTest.java
jzhao-talend: @sgandon 
ok. I would fix it. Thanks.

components-common/src/main/java/org/talend/components/common/ProxyProperties.java
francisu: But it seems there might be an NPE problem (that you have already found), can you track that down?

PR #134, opened 2016-06-02
TCOMP-190 Implement main and reject flows for output components.

RyanSkraba:


components-salesforce/src/test/java/org/talend/components/salesforce/runtime/SalesforceSinkTestIT.java
```diff
@@ -0,0 +1,377 @@
+// ============================================================================
```
sgandon: I would suggest you move those test in the WriterTest class instead of the Sinnk test ;)

components-salesforce/src/test/java/org/talend/components/salesforce/runtime/SalesforceWriterTestIT.java
```diff
@@ -146,80 +267,246 @@ protected void runOutputInsert(boolean isDynamic) throws Exception {
...
+        assertThat(sfWriter.getRejectedWrites(), hasSize(1));
+
+        // Check the rejected record.
+        IndexedRecord rejected = sfWriter.getRejectedWrites().get(0);
+        assertThat(rejected.getSchema().getFields(), hasSize(7));
+
+        // Check the values copied from the incoming record.
+        for (int i = 0; i < r.getSchema().getFields().size(); i++) {
+            assertThat(rejected.getSchema().getFields().get(i), is(r.getSchema().getFields().get(i)));
+            assertThat(rejected.get(0), is(r.get(0)));
```
wwang-talend: assertThat(rejected.get(0), is(r.get(0)));

should : get(i)?

RyanSkraba: Yes, thanks!  I will correct and verify.

PR #122, opened 2016-05-25
Feature/tdi 35967 jira output component

gonchar-ivan:
Implemented Jira Output component.
https://jira.talendforge.org/browse/TDI-35967

PR #138, opened 2016-06-03
Bugfix/tdi 36417 fix jira input properties

gonchar-ivan:
Fix refresh properties
Fix return properties
for Jira components

Refer to:
https://jira.talendforge.org/browse/TDI-36416
https://jira.talendforge.org/browse/TDI-36444

### DIFF
--- a/components-jira/src/main/java/org/talend/components/jira/JiraConnectionProperties.java
+++ b/components-jira/src/main/java/org/talend/components/jira/JiraConnectionProperties.java
@@ -12,8 +12,8 @@
 // ============================================================================
 package org.talend.components.jira;
 
-import org.talend.components.api.properties.ComponentPropertiesImpl;
 import org.talend.daikon.properties.Properties;
+import org.talend.daikon.properties.PropertiesImpl;
 import org.talend.daikon.properties.presentation.Form;
 import org.talend.daikon.properties.property.PropertyFactory;
 import org.talend.daikon.properties.property.StringProperty;
@@ -21,7 +21,7 @@ import org.talend.daikon.properties.property.StringProperty;
 /**
  * Jira connection {@link Properties}
  */
-public class JiraConnectionProperties extends ComponentPropertiesImpl {
+public class JiraConnectionProperties extends PropertiesImpl {
 
     private static final String INITIAL_HOST = "https://jira.atlassian.com";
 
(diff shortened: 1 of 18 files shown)


components-jira/src/main/java/org/talend/components/jira/JiraConnectionProperties.java
```diff
@@ -12,16 +12,16 @@
...
+import org.talend.daikon.properties.PropertiesImpl;
 import org.talend.daikon.properties.presentation.Form;
 import org.talend.daikon.properties.property.PropertyFactory;
 import org.talend.daikon.properties.property.StringProperty;
 
 /**
  * Jira connection {@link Properties}
  */
-public class JiraConnectionProperties extends ComponentPropertiesImpl {
+public class JiraConnectionProperties extends PropertiesImpl {
```
wwang-talend: why change from ComponentPropertiesImpl to PropertiesImpl ? only want to know...

gonchar-ivan: 1. As for me ComponentPropertiesImpl should be extended only by root properties. Here, JiraConnectionProperties is nested properties of JiraProperties.
2. ComponentPropertiesImpl contains "returns" property, which have no meaning for JiraConnectionProperties and other nested properties

gonchar-ivan: also for reference my discussion with Sebastien in Service chat (01.06 Wednesday):
[6/1/2016 5:00:42 PM] iv.gonchar: Why all Properties should extend from ComponentProperties?
For example, I am implementing JiraConnectionProperties, which is a part of JiraInputProperties and JiraOutputProperties.
I tried to extend JiraConnectionProperties from Properties, but got ClassCastException in Studio, because of following code:

https://github.com/Talend/tdi-studio-se/blob/1f34e0691fe620253b112df03b433b1a4019ad4e/main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/utils/ComponentsUtils.java#L194

ComponentProperties contain methods and fields, which are not requeired for JiraConnectionProperties. Such as following:

Property<String> returns
getSchema()
getPossibleConnectors()
...
[6/1/2016 5:55:22 PM] iv.gonchar: Fixed Jira components according review comments:
https://github.com/Talend/components/pull/122/

Please, check it again.
[6/1/2016 6:08:33 PM] Sébastien Gandon: @Ivan, you are absolutely right about the Properties vs ComponentProperties but it has too much an impact right now, for instance the services callbacks should be working on Properties and not ComponentProperties. Please create a jira on TCOMP for this. This is something I also thought about about.

gonchar-ivan: https://jira.talendforge.org/browse/TCOMP-201

wwang-talend:
now it have a conflict to master, please fix it.

PR #139, opened 2016-06-03
TCOMP-202 Added display names for common returns properties

gonchar-ivan:
WIP

https://jira.talendforge.org/browse/TCOMP-202

Also return properties display names  should be removed from all components. But currently I can't check it in Studio, because of old component-api version in Studio (0.11.0)


sgandon:
this PR is fine by me although I don't like having those DI specific names in the ComponentProperties Impl I am not sure where else we could place them to be easily available. May be some Constant class ?

PR #143, opened 2016-06-06
TDI-36051: Support dynamic schema handling

stavytskyi:


components-dataprep/src/main/java/org/talend/components/dataprep/runtime/DataPrepAdaptorFactory.java
```diff
@@ -39,6 +48,11 @@ public Schema getSchema() {
 
     @Override
     public IndexedRecord convertToAvro(DataPrepField[] dataPrepDataSetRecord) {
+        if (firstRow && AvroUtils.isIncludeAllFields(schema)) {
```
wwang-talend: firstRow is true at anytime, is it right?

stavytskyi: no, It's my misspelling, I'll update commit and remove. It was left after removing not needed a code.

wwang-talend:
now it have a conflict to master, please fix it.

PR #151, opened 2016-06-08
TDKN-26 Refactor and rename Avro packages.

RyanSkraba:
See https://github.com/Talend/daikon/pull/44


sgandon:
please merge it as soon as possible this is just renames following some refactor.

PR #152, opened 2016-06-10
TCOMP-165 Unify return value handling

francisu:
This change removes the return values from ComponentProperties and instead adds their declarations to ComponentDefinition. It also changes the runtime to provide the return values as a Map<String, Object> in Reader.getReturnValues() and as the return for WriterOperation.finalize().

The names of the return values have also changed from being DI-specific to something more clear and fitting with the framework. See ComponentDefinition for the new names. The DI code will need to map these to the DI names, i.e. totalRecordCount -> NB_LINE.

Finally the WriterResult has been replaced by Result which is used both by readers and writers to capture the return values.


components-api/src/main/java/org/talend/components/api/component/runtime/Reader.java
```diff
@@ -120,11 +130,19 @@
...
     Source getCurrentSource();
+
+    /**
+     * Gets the return values from the complete operation of the reader.
+     *
+     * See {@link ComponentDefinition#getReturnProperties()}, also {@link Result}.
+     *
+     * @return a map of the return values.
+     */
+    Map<String, Object> getReturnValues();
```
gonchar-ivan: Why it returns Map<String, Object>, but not Map<String, Result>  ?

components-api/src/main/java/org/talend/components/api/component/runtime/Result.java
```diff
@@ -0,0 +1,131 @@
...
+        map.put(ComponentDefinition.RETURN_SUCCESS_RECORD_COUNT, successCount);
+        map.put(ComponentDefinition.RETURN_REJECT_RECORD_COUNT, rejectCount);
+        return map;
+    }
+
+    public Map<String, Object> toMap() {
+        return createResultMap(totalCount, successCount, rejectCount);
+    }
+
+    public int getTotalCount() {
```
gonchar-ivan: Fields are public. As a rule getters return fields without any additional logic.
Maybe getters are redundant here.

components-jira/src/main/java/org/talend/components/jira/runtime/reader/JiraReader.java
```diff
@@ -100,6 +101,8 @@
      */
     private boolean hasMoreRecords;
 
+    private Result result;
```
gonchar-ivan: I don't like it as a field.
It could be moved to getReturnValues() method.

But if the idea is to make Result class fields public, then I agree.
It this case I don't need entityCounter field. I can count records as follows
result.totalCount++

components-jira/src/main/java/org/talend/components/jira/runtime/reader/JiraReader.java
```diff
@@ -214,6 +212,17 @@ public Source getCurrentSource() {
...
     }
 
+
+    @Override
+    public void close() throws IOException {
+    }
+
+    @Override
+    public Map<String, Object> getReturnValues() {
+        result.totalCount = entityCounter;
```
gonchar-ivan: It could be:
Result result = new Result("uid???", entityCounter);
return result.toMap();

components-jira/src/main/java/org/talend/components/jira/runtime/reader/JiraReader.java
```diff
@@ -12,18 +12,12 @@
 // ============================================================================
 package org.talend.components.jira.runtime.reader;
 
-import java.io.IOException;
```
gonchar-ivan: What rule for organize imports do you use?
I am using:
...\components\tooling\IDEs\eclipse\talend.importorder

Its content:
#Organize Import Order
#Tue Mar 24 10:40:36 CET 2015
3=com
2=org
1=javax
0=java

Here I had imports according above rule.
Do I use wrong .importorder file?

components-jira/src/main/java/org/talend/components/jira/tjirainput/TJiraInputDefinition.java
```diff
@@ -59,5 +61,9 @@ public Source getRuntime() {
     public Class<? extends ComponentProperties> getPropertyClass() {
         return TJiraInputProperties.class;
     }
-    
+
+    @Override
+    public Property[] getReturnProperties() {
+        return new Property[] { newProperty(RETURN_TOTAL_RECORD_COUNT) };
```
gonchar-ivan: What about type of return property (String, Integer, etc)?
Does Studio (or other Talend product) require it?

components-jira/src/main/java/org/talend/components/jira/runtime/JiraWriteOperation.java
```diff
@@ -80,47 +69,36 @@ public JiraWriter createWriter(RuntimeContainer container) {
...
-            return new JiraInsertWriter(this);
-        }
-        case UPDATE: {
-            return new JiraUpdateWriter(this);
-        }
-        default: {
-            LOG.debug("Impossible action retrieved from Jira sink");
-            return null;
-        }
+            case DELETE: {
```
gonchar-ivan: Current formatting rules in
tools\tools-java-formatter\src\main\resources\talend_java_eclipse_formatter.xml
contains
<setting id="org.eclipse.jdt.core.formatter.indent_switchstatements_compare_to_switch" value="false"/>

but seems your formatter has true value for switchstatements indent.
What should be used?

components-api/src/main/java/org/talend/components/api/component/ComponentDefinition.java
```diff
@@ -48,14 +51,35 @@
     public Trigger[] getTriggers();
 
     /**
+     * Common return properties names
+     */
+    public static final String RETURN_ERROR_MESSAGE = "errorMessage";
+
+    public static final String RETURN_TOTAL_RECORD_COUNT = "totalRecordCount";
```
gonchar-ivan: For now Studio displays return values as 
{displayName} - {Name}
E.g.
Number of lines - NB_LINE

After this change it will be like
Number of lines - totalRecordCount

Is it correct?

gonchar-ivan: Also
ComponentDefinition.properties

has other properties names

components-api/src/main/java/org/talend/components/api/component/runtime/Result.java
francisu: Probably so, I made the fields public so that you could just use them directly if you wanted. When I adjusted the components, I think I did it different ways for each component.

components-jira/src/main/java/org/talend/components/jira/runtime/reader/JiraReader.java
francisu: Yeah, could do it a lot of ways (and as I said before I did it differently I think in the Jira component than the others). I think the way I would prefer is to have the result be a field and then create it at open time and not even keep the uId as a field. And update the counters in just the result. I will go ahead and make that change, and if you want something else you can fix it later if that's OK.

components-jira/src/main/java/org/talend/components/jira/runtime/reader/JiraReader.java
francisu: This is some problem in my intelliJ configuration. I will fix this.

components-jira/src/main/java/org/talend/components/jira/tjirainput/TJiraInputDefinition.java
francisu: Return could be anything, it should match what's declared in the Property class, but there is no checking for that. (The previous mechanism had the same issue).

components-api/src/main/java/org/talend/components/api/component/ComponentDefinition.java
francisu: Yes, as I mentioned in my opening comment, if the studio wants to keep the old names like NB_LINE (for the properties defined in ComponentDefinition), then it can map them. Or they can just use the new names. The way of handling property names needs to be consistent for the TCOMP framework which has to support things beyond the studio/DI use, which is why I'm making this change. For returns that are not defined in ComponentDefinition, those will follow the new camel case convention. The studio could convert it automatically to uppercase and use underscores where the case changes. I will talk with the studio team about this.

PR #154, opened 2016-06-10
TDP-2082: Adds ?tag=components to create dataset URL

fhuaulme:
* Adds &tag=components to create URL
* Adds unit test assertions for component URL (+ additions on mock server to actually check last request values).


francisu:
This is fine, but why are we using a Mock for the server? We already have one of those, we should be using the real thing.

PR #159, opened 2016-06-14
TUP-4570: fix the long type converter issue

wwang-talend:
https://jira.talendforge.org/browse/TUP-4570


francisu:
I'm not seeing that these changes are being reviewed by others, and I don't see a junit test for this bug. Every fix needs to have a junit test that shows the problem.

wwang-talend:
The issue is caused by the empty string pattern for long type, it's about the ui model value of column pattern in talend ui schema config. We can add a junit to test it by only setting a fake empty string pattern if it's necessay. And now we only test it on component-0.12 and daikon-0.11 as https://jira.talendforge.org/browse/TUP-4851 make it difficult to test studio and component together. We will add the junit after TUP-4851's ok.

francisu:
You don't need to worry about adding a unit test for this in the studio. You should add a unit test here in components for any fix you make in components. You have to keep in mind the new components will be used by things other than the studio. The problem with TUP-4851 will not prevent you from adding a unit test here.

wwang-talend:
https://github.com/Talend/components/pull/161
Have added the case, please merge it.

PR #161, opened 2016-06-15
TUP-4570: add a junit for the long type with empty string pattern

wwang-talend:
https://jira.talendforge.org/browse/TUP-4570

Conflicts:
	components-common/src/test/java/org/talend/components/common/runtime/BulkFileWriterTest.java


francisu:
Thanks very much!

PR #163, opened 2016-06-17
Bug/6.2/tup 4570 type issues

wwang-talend:


components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkFileWriter.java
```diff
@@ -60,8 +60,14 @@ public SalesforceBulkFileWriter(WriteOperation<Result> writeOperation, BulkFileP
...
-                        if (polymorphics.get(index)) {
+                        
+                        Object polymorphic = polymorphics.get(index);
+                        boolean poly = false;
+                        if(polymorphic!=null && polymorphic instanceof Boolean) {
+                            poly = (Boolean)polymorphic;
+                        }
+                        
+                        if (poly) {
                             sbuilder.append(lookupFieldModuleNames.get(index)).append(":");
```
francisu: Formatting is not right here. Can you check your formatter?

PR #164, opened 2016-06-17
TUP-4585: Backport TDI-35282 on new Framework

jzhao-talend:
https://jira.talendforge.org/browse/TUP-4585
Note:Support lookup field null value update

### DIFF
--- a/components-salesforce/src/main/java/org/talend/components/salesforce/UpsertRelationTable.java
+++ b/components-salesforce/src/main/java/org/talend/components/salesforce/UpsertRelationTable.java
@@ -1,6 +1,6 @@
 package org.talend.components.salesforce;
 
-import static org.talend.daikon.properties.property.PropertyFactory.*;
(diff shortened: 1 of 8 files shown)


jzhao-talend:
Need merge together with https://github.com/Talend/tdi-studio-se/pull/494

jzhao-talend:
I can't find way to add junit test the runtime of upsert relationship table with default modules.
So only add a talend test job (export from 6.1.1)
https://jira.talendforge.org/secure/attachment/99835/TestCase_tSalesforceOutput_upsert_relationship.zip

francisu:
@jzhao-talend what's the issue with being able to test it with Junit? We need to have all of the tests be done through the Junit mechanism here. We cannot rely on the studio tests.

jzhao-talend:
@francisu 
yes. studio test is only for component developer and QA. Please ignore that.
My studio job is work with a customer module which was created in my salesforce account.
I don't know the relationship mapping with the default modules with default fields(Account/User/...)

PR #165, opened 2016-06-17
TDI-36463: tSalesforceConnection - 'Security Key' should NOT be a required parameter

jzhao-talend:
https://jira.talendforge.org/browse/TDI-36463


jzhao-talend:
Fixed the required issues all in components side.
So https://jira.talendforge.org/browse/TDKN-54 can be changed to minor one.

components-common/src/main/java/org/talend/components/common/ProxyProperties.java
```diff
@@ -46,6 +46,13 @@ public ProxyProperties(String name) {
...
 
     @Override
+    public void setupProperties() {
+        super.setupProperties();
+        userPassword.userId.setOccurMinTimes(0);
+        userPassword.password.setOccurMinTimes(0);
+    }
+
+    @Override
     public void setupLayout() {
```
francisu: You don't need this code. By default properties occur 0 min times, so it should not be specified here.

jzhao-talend: @francisu 
Thanks for you response.
Yes. the default min occur is 0, but here use a common Properties: [UserPasswordProperties.java](https://github.com/Talend/components/blob/master/components-common/src/main/java/org/talend/components/common/UserPasswordProperties.java)
The "userId" and "password" all init as requred. 
In ProxyProperties.java, they are not always required. but we can't change it to not required.
That's why I created [TDKN-54](https://jira.talendforge.org/browse/TDKN-54)

francisu: Yes, got it. And actually I fixed TDKN-54 by removing the boolean (which was probably not what you intended -- sorry about that). We should add the setRequired(boolean) back and have it change the numbers to 0/1. That would make this a little cleaner. Would you mind just adding that method back? You can do it as part of this PR if you like.

jzhao-talend: ok. I would add that method back.

jzhao-talend: Created the pull request:
https://github.com/Talend/daikon/pull/47

PR #170, opened 2016-06-21
TDI-36303: Change  RuntimeException to ComponentException.

jzhao-talend:
https://jira.talendforge.org/browse/TDI-36303


components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceAvroRegistry.java
```diff
@@ -284,7 +285,7 @@ public Long convertToAvro(String value) {
             } catch (ParseException e) {
                 // TODO: error handling
                 e.printStackTrace();
```
sgandon: no printStackTrace() ever please.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceWriter.java
```diff
@@ -454,7 +455,7 @@ else if (TSalesforceOutputProperties.FIELD_ERROR_FIELDS.equals(outField.name()))
             String ID = "Id";
             Schema.Field idField = input.getSchema().getField(ID);
             if (null == idField)
-                throw new RuntimeException(ID + " not found");
+                throw new ComponentException(new Throwable(ID + " not found"));
```
sgandon: please suround this with brackets {}

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
```diff
@@ -421,7 +396,7 @@ public void doBulkQuery(String moduleName, String queryStatement, int secToWait)
                 queryResultIDs = new HashSet<String>(Arrays.asList(list.getResult())).iterator();
                 break;
             } else if (info.getState() == BatchStateEnum.Failed) {
-                throw new RuntimeException("-------------- failed ----------" + info);
+                throw new ComponentException(new Throwable("-------------- failed ----------" + info));
```
sgandon: I don't like the fact you create a Throwable for some expected error. I would advise you to create one specific ErrorCode for this case and use it with a context, see this for example.
https://github.com/Talend/components/blob/master/components-common/src/main/java/org/talend/components/common/FixedConnectorsComponentProperties.java#L63
I am working at making it more simpler to setup context but meanwhile you can do like the example.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceWriter.java
sgandon: same comment as above, please create some specific ErrorCode.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
jzhao-talend: @sgandon 
Thanks for your checking.
So I need to create enum like this?
`public enum SalesforceRuntimeErrorCode implements ErrorCode {
    EXECUTE_BATCH_FAILED(HttpServletResponse.SC_BAD_REQUEST, "runtime"), 
    SF_ID_MISSING(HttpServletResponse.SC_BAD_REQUEST, "runtime");
//....
}`

PR #171, opened 2016-06-21
TDI-36651: adjust the return property for salesforce components

wwang-talend:
https://jira.talendforge.org/browse/TDI-36651

### DIFF
--- a/components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceDefinition.java
+++ b/components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceDefinition.java
@@ -12,15 +12,15 @@
 // ============================================================================
 package org.talend.components.salesforce;
 
+import static org.talend.daikon.properties.property.PropertyFactory.newInteger;
+import static org.talend.daikon.properties.property.PropertyFactory.newProperty;
+
 import org.talend.components.api.component.AbstractComponentDefinition;
 import org.talend.components.api.properties.ComponentProperties;
 import org.talend.daikon.properties.property.Property;
 
-import static org.talend.daikon.properties.property.PropertyFactory.newProperty;
-import static org.talend.daikon.properties.property.PropertyFactory.newInteger;
-
 public abstract class SalesforceDefinition extends AbstractComponentDefinition {
-    
+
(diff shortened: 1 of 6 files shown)


components-salesforce/src/main/java/org/talend/components/salesforce/SalesforceDefinition.java
```diff
@@ -39,7 +39,7 @@ public SalesforceDefinition(String componentName) {
     @Override
     // Most of the components are on the input side, so put this here, the output definition will override this
     public Property[] getReturnProperties() {
-        return new Property[] { newInteger(RETURN_TOTAL_RECORD_COUNT) };
+        return new Property[] { newProperty(RETURN_ERROR_MESSAGE), newInteger(RETURN_TOTAL_RECORD_COUNT) };
```
gonchar-ivan: I am not sure we need to add ERROR_MESSAGE to components

wwang-talend: The part also affect the ui part, so it's necessary now.

wwang-talend: Please see the outline view in talend studio for salesforce components, it affect that.

wwang-talend: In future, we can remove it and make it in framework part, but now just make it work.

gonchar-ivan: Wang Wei, thank you very much for explanation.
Do you know what class in Studio code is responsible for outline view?

wwang-talend: Please see this : org.talend.designer.core.generic.model.Component.createReturns()

sgandon: I hate to see the same returns for every components, if all components need a ERROR_MESSAGE return then we can remove them all from components and change the studio to consider that all of the component have them.

PR #172, opened 2016-06-21
TCOMP-217 : added I18N for commons Returns values

sgandon:
There is not way to setup I18N if you create local variable Property instances.
So in order to easy the component developer I have created the usual returns properties as static instances and i18n is initialized in the AbsctractComponentDefinition constructor.
For specific returns properties the user will have to setup i18n by way of code.


components-api-service/src/test/java/org/talend/components/api/test/ComponentTestUtils.java
```diff
@@ -58,8 +59,18 @@ static public void testAlli18n(ComponentService componentService, ErrorCollector
...
-            // check component properties title
+            // check component definition title
             errorCollector.checkThat("missing I18n property :" + cd.getTitle(), cd.getTitle().contains("component."), is(false));
+            // check return properties i18n
+            checkAllPropertyI18n(cd.getReturnProperties(), cd, errorCollector);
+        }
+    }
+
+    public static void checkAllPropertyI18n(Property[] propertys, Object parent, ErrorCollector errorCollector) {
+        if (propertys != null) {
```
gonchar-ivan: mispelled
propertIEs, ;-)

sgandon: I did it on purpose to differentiate Property instance with Properties instance, but I'll change it back you are probably right.

components-api/src/main/resources/org/talend/components/api/component/AbstractComponentDefinition.properties
```diff
@@ -0,0 +1,4 @@
+property.errorMessage.displayName=Error Message
```
gonchar-ivan: I have a question about it.
Does i18n mechanism looks for some key in parent class .properties file, if current .properties file doesn't contain that key? 
E.g. if AbstractComponentDefinition.properties missed some key. Will i18n mechanism looks for it in ComponentDefinition.properties?

components-api/src/test/java/org/talend/components/api/service/testcomponent/TestComponentDefinition.java
```diff
@@ -43,9 +44,11 @@ public TestComponentDefinition() {
 
     @Override
     public Property[] getReturnProperties() {
-        return new Property[] { newProperty("return1"), newProperty(RETURN_ERROR_MESSAGE),
-                newInteger(RETURN_TOTAL_RECORD_COUNT), newInteger(RETURN_SUCCESS_RECORD_COUNT),
-                newInteger(RETURN_REJECT_RECORD_COUNT) };
+        StringProperty return1 = newProperty("return1");
+        setupI18N(new Property<?>[] { return1 });
+
+        return new Property[] { return1, RETURN_ERROR_MESSAGE_PROP, RETURN_TOTAL_RECORD_COUNT_PROP,
```
gonchar-ivan: What do you think about following possible getReturnProperties() implementation?:

    Property[] returnProperties = new Property[] {newProperty(RETURN_ERROR_MESSAGE), newInteger(RETURN_TOTAL_RECORD_COUNT), newInteger(RETURN_SUCCESS_RECORD_COUNT), newInteger(RETURN_REJECT_RECORD_COUNT)};

    return setupI18N(returnProperties);
}

In this case there is no need to static return properties in AbstractComponentDefinition.

components-splunk/src/main/resources/org/talend/components/splunk/messages.properties
```diff
@@ -2,4 +2,5 @@ component.tSplunkEventCollector.title=Splunk Event Collector
 form.Main.title=Main
 form.Advanced.title=Advanced
 form.Reference.title=Reference
-property.returns.displayName=returns
\ No newline at end of file
+property.returns.displayName=returns
```
gonchar-ivan: Seems, 
property.returns.displayName is not needed

components-api-service/src/test/java/org/talend/components/api/test/ComponentTestUtils.java
gonchar-ivan: Maybe
propertyList or propertyArray 
will be suitable for this case

components-api/src/main/resources/org/talend/components/api/component/AbstractComponentDefinition.properties
sgandon: Yes you are right it only look for parent class and not interfaces. If you consider it could be improved, please create a Jira or even better create a PR in daikon for it.

components-api/src/test/java/org/talend/components/api/service/testcomponent/TestComponentDefinition.java
sgandon: That is what I did first but it is much more easier for the developer to just use a static instance which make sense because we already have static strings for those expected returns. More over this avoid creating multiple instances of the same thing.

components-splunk/src/main/resources/org/talend/components/splunk/messages.properties
sgandon: right I'll remove it.

components-api-service/src/test/java/org/talend/components/api/test/ComponentTestUtils.java
sgandon: I agree.

components-api/src/test/java/org/talend/components/api/service/testcomponent/TestComponentDefinition.java
gonchar-ivan: Thank you.
Now I understand.

gonchar-ivan:
Ok for me

PR #174, opened 2016-06-22
TDI-36474: tDatasetOutput doesn't have any message to hint job fail

stavytskyi:


stavytskyi:
This pull request contains fix related to: https://github.com/Talend/tdi-studio-se/pull/509

PR #176, opened 2016-06-23
TDI-36711: added check if writer was closed earlier, added success an…

chmyga:
…d reject count values for writer.

### DIFF
--- a/components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
+++ b/components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
@@ -135,7 +135,7 @@ public class TSplunkEventCollectorWriter implements Writer<Result> {
         String jsonResponseString = EntityUtils.toString(response.getEntity());
         try {
             handleResponse(jsonResponseString);
-        } catch (Exception e) {
+        } catch (IOException e) {
             rejectCount += splunkObjectsForBulk.size();
             throw e;
         } finally {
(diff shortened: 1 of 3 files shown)


components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
```diff
@@ -72,6 +77,8 @@
 
     private I18nMessages messageFormatter;
 
+    private AtomicBoolean closed = new AtomicBoolean(false);
```
sgandon: Can you explain wy you used an AtomicBoolean here ?

components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
```diff
@@ -119,13 +127,17 @@ private void doSend() throws IOException {
...
         }
+        dataCount += splunkObjectsForBulk.size();
         HttpPost request = createRequest(splunkObjectsForBulk);
 
         HttpResponse response = splunkConnection.sendRequest(request);
 
         String jsonResponseString = EntityUtils.toString(response.getEntity());
         try {
             handleResponse(jsonResponseString);
+        } catch (Exception e) {
```
sgandon: could you only catch IOEXception please

components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
chmyga: Just in case there are several threads which would like to close the same writer. Though I can change it to primitive boolean value.

components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
chmyga: yes, sure. I will fix it.

components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
sgandon: Ok that is fine, please leave it.
Thanks.

PR #182, opened 2016-06-24
TDI-36474: tDatasetOutput doesn't have any message to hint  job fail

wwang-talend:
https://jira.talendforge.org/browse/TDI-36474


sgandon:
A Unit test for this fix should be provided.

PR #184, opened 2016-06-26
TDI-36685: improve the error information

wwang-talend:
https://jira.talendforge.org/browse/TDI-36685

### DIFF
--- a/components-dataprep/src/main/java/org/talend/components/dataprep/connection/DataPrepConnectionHandler.java
+++ b/components-dataprep/src/main/java/org/talend/components/dataprep/connection/DataPrepConnectionHandler.java
@@ -57,16 +57,19 @@ public class DataPrepConnectionHandler {
 
     private final String dataSetName;
 
+    private final String dataSetId;
+
     private DataPrepOutputModes mode;
 
     private HttpURLConnection urlConnection;
 
     private Header authorisationHeader;
 
-    public DataPrepConnectionHandler(String url, String login, String pass, String dataSetName) {
+    public DataPrepConnectionHandler(String url, String login, String pass, String dataSetId, String dataSetName) {
         this.url = url;
         this.login = login;
         this.pass = pass;
+        this.dataSetId = dataSetId;
         this.dataSetName = dataSetName;
         LOGGER.debug("Url: {}", url);
     }
@@ -116,13 +119,13 @@ public class DataPrepConnectionHandler {
             switch (mode) {
             case Create:
                 urlConnection.disconnect();
-                throw new IOException("Dataset exist on Dataprep server. Response code: " + errorMessage);
+                throw new IOException("Dataset exist on Dataprep server. Response information: " + errorMessage);
             case Update:
                 urlConnection.disconnect();
-                throw new IOException("Wrong DatasetID. Response code: " + errorMessage);
+                throw new IOException("Wrong DatasetID. Response information: " + errorMessage);
             case LiveDataset:
                 urlConnection.disconnect();
-                throw new IOException("Wrong url. Response code: " + errorMessage);
+                throw new IOException("Wrong url. Response information: " + errorMessage);
             default:
                 throw new UnsupportedOperationException();
             }
@@ -146,7 +149,7 @@ public class DataPrepConnectionHandler {
     }
 
     public DataPrepStreamMapper readDataSetIterator() throws IOException {
-        Request request = Request.Get(url + API_DATASETS + dataSetName + "?metadata=false").addHeader(authorisationHeader);
+        Request request = Request.Get(url + API_DATASETS + dataSetId + "?metadata=false").addHeader(authorisationHeader);
         HttpResponse response = request.execute().returnResponse();
         if (returnStatusCode(response) != HttpServletResponse.SC_OK) {
             String moreInformation = extractResponseInformationAndConsumeResponse(response);
@@ -163,7 +166,7 @@ public class DataPrepConnectionHandler {
         case Create:
             return writeToServer("POST", requestEncoding());
         case Update:
-            return writeToServer("PUT", url + API_DATASETS + dataSetName);
+            return writeToServer("PUT", url + API_DATASETS + dataSetId);
         case LiveDataset:
             return writeToServer("POST", url);
         default:
@@ -199,7 +202,7 @@ public class DataPrepConnectionHandler {
     }
 
     public List<Column> readSourceSchema() throws IOException {
-        Request request = Request.Get(url + API_DATASETS + dataSetName + "/metadata");
+        Request request = Request.Get(url + API_DATASETS + dataSetId + "/metadata");
         request.addHeader(authorisationHeader);
 
         DataPrepStreamMapper dataPrepStreamMapper = null;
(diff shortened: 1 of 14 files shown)


components-dataprep/src/main/java/org/talend/components/dataprep/connection/DataPrepConnectionHandler.java
```diff
@@ -213,4 +225,39 @@ private OutputStream writeToServer(String requestMethod, String request) throws
 
         return metaData.getColumns();
     }
+
+    private String extractResponseInformationAndConsumeResponse(HttpResponse response) throws IllegalStateException, IOException {
+        InputStream is = response.getEntity().getContent();
+        return extractResponseInformationAndConsumeResponse(is);
```
sgandon: I would suggest a try/finally block here to close the input stream that you created.
It is usually better to have the code that closes the stream at the same level as the one opening it.

components-dataprep/src/main/java/org/talend/components/dataprep/connection/DataPrepConnectionHandler.java
```diff
@@ -213,4 +225,39 @@ private OutputStream writeToServer(String requestMethod, String request) throws
...
+    }
+
+    private String extractResponseInformationAndConsumeResponse(HttpURLConnection connection) throws IOException {
+        InputStream is = null;
+        try {
+            is = connection.getInputStream();
+        } catch (IOException e) {
+            is = connection.getErrorStream();
+        }
+        return extractResponseInformationAndConsumeResponse(is);
```
sgandon: same comment here.

components-dataprep/src/main/java/org/talend/components/dataprep/tdatasetinput/TDataSetInputProperties.java
```diff
@@ -78,8 +80,19 @@ public TDataSetInputProperties(String name) {
...
         Form form = getForm(Form.MAIN);
-        form.addRow(dataSetName);
+        form.addRow(Widget.widget(this.dataSetName).setWidgetType("widget.type.dataset.selection"));
         form.addRow(Widget.widget(fetchSchema).setWidgetType(Widget.BUTTON_WIDGET_TYPE));
+
+        Form advancedForm = new Form(this, Form.ADVANCED);
+        advancedForm.addRow(dataSetId);
+    }
+
+    public static final String ADD_QUOTES = "ADD_QUOTES";
```
sgandon: What is the reason for the ADD_QUOTES ?
In the component there should not be any knowledge of the Studio and therefore the idea that the Studio uses quotes or not for a value is not of the concern of the Component Developer.

components-dataprep/src/main/java/org/talend/components/dataprep/tdatasetoutput/TDataSetOutputProperties.java
```diff
@@ -132,11 +180,62 @@ public RuntimeProperties getRuntimeProperties() {
...
+            runtimeProperties.setDataSetName(dataSetNameForCreateMode.getStringValue());
         } else {
-            runtimeProperties.setDataSetName(dataSetId.getStringValue());
+            runtimeProperties.setDataSetId(dataSetId.getStringValue());
         }
         runtimeProperties.setLimit(limit.getStringValue());
         return runtimeProperties;
     }
+
+    public ValidationResult afterFetchSchema() {
```
sgandon: I don't see any unit test for this method, please provide one.

components-dataprep/src/main/java/org/talend/components/dataprep/tdatasetinput/TDataSetInputProperties.java
wwang-talend: dataSetName is a special field without double quote, and can't set it to hidden. So when generate the code, a compile issue appear. Avoid this issue by this code.

components-dataprep/src/main/java/org/talend/components/dataprep/tdatasetoutput/TDataSetOutputProperties.java
wwang-talend: have added one.

components-dataprep/src/main/java/org/talend/components/dataprep/connection/DataPrepConnectionHandler.java
wwang-talend: thanks for your advice, have fixed it.

components-dataprep/src/main/java/org/talend/components/dataprep/tdatasetinput/TDataSetInputProperties.java
sgandon: what does it make it special, all String field do not have quotes anyway from the Component developer point of view.

components-dataprep/src/test/java/org/talend/components/dataprep/tdatasetinput/AfterFetchSchemaTest.java
```diff
@@ -35,21 +36,35 @@
...
+    // TODO should remove the duplicated code for afterFetchSchema
+    @Test
+    public void testAfterFetchSchemaForOutput() {
+        TDataSetOutputProperties properties = (TDataSetOutputProperties) componentService.getComponentDefinition("tDatasetOutput")
+                .createProperties();
+        properties.url.setValue("http://localhost:" + serverPort);
+        properties.login.setValue("vincent@dataprep.com");
+        properties.pass.setValue("vincent");
+        properties.dataSetName.setValue("mydataset");
+        properties.dataSetId.setValue("db119c7d-33fd-46f5-9bdc-1e8cf54d4d1e");
```
sgandon: Please provide also test cases for the different possible errors

components-dataprep/src/main/java/org/talend/components/dataprep/connection/DataPrepConnectionHandler.java
```diff
@@ -227,18 +227,31 @@ private OutputStream writeToServer(String requestMethod, String request) throws
     }
 
     private String extractResponseInformationAndConsumeResponse(HttpResponse response) throws IllegalStateException, IOException {
-        InputStream is = response.getEntity().getContent();
-        return extractResponseInformationAndConsumeResponse(is);
+        InputStream is = null;
```
sgandon: I would suggest this.
```
InputStream is = response.getEntity().getContent();
 try {
       return extractResponseInformationAndConsumeResponse(is);
} finally {
       is.close();
}
```

components-dataprep/src/test/java/org/talend/components/dataprep/tdatasetinput/AfterFetchSchemaTest.java
wwang-talend: get it, will do it, but before creating the cases, i need to move afterFetchSchema to the parent class from TDataSetOutputProperties and TDataSetInputProperties.

components-dataprep/src/main/java/org/talend/components/dataprep/connection/DataPrepConnectionHandler.java
wwang-talend: agree, to do it.

PR #185, opened 2016-06-27
TDI-36730: Live dataset on MySQL data all shifted in data prep (due to

wwang-talend:
nulls?)
https://jira.talendforge.org/browse/TDI-36730


francisu:
Why was there no unit test for this?

PR #186, opened 2016-06-27
TDI-36536: added parsing of String Dates to Dates using the design-ti…

chmyga:
…me pattern.

### DIFF
--- a/components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
+++ b/components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
@@ -19,6 +19,10 @@ import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.avro.Schema;
+import org.apache.avro.Schema.Field;
+import org.apache.avro.SchemaBuilder;
+import org.apache.avro.SchemaBuilder.FieldAssembler;
+import org.apache.avro.SchemaBuilder.FieldBuilder;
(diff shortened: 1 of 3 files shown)


components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
```diff
@@ -79,13 +87,49 @@
...
+
+    private void setSchema(Schema designSchema) {
+        if (designSchema != null && AvroUtils.isIncludeAllFields(designSchema)) {
+            this.defaultSchema = initDefaultSchema(designSchema);
+        } else {
+            this.defaultSchema = designSchema;
+        }
+    }
+
+    private Schema initDefaultSchema(Schema designSchema) {
```
sgandon: please add Unit tests for this.

components-splunk/src/main/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriter.java
```diff
@@ -111,8 +155,17 @@ public void write(Object datum) throws IOException {
         SplunkJSONEvent event = SplunkJSONEventBuilder.createEvent();
 
         for (Schema.Field f : input.getSchema().getFields()) {
-            if (input.get(f.pos()) != null) {
-                SplunkJSONEventBuilder.setField(event, f.name(), input.get(f.pos()), true);
+            Schema.Field defaultField = null;
```
sgandon: and unit test for this as well.

francisu:
Why was this merged when the comments were not addressed?

PR #191, opened 2016-06-29
TDI-36536: added integration tests for dynamic time column.

chmyga:


build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/23/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/23/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/27/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/27/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/31/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/31/)

components-splunk/src/test/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriterTestIT.java
```diff
@@ -157,6 +197,14 @@ private Schema createSchema() {
...
 
+    private Schema createDynamicSchemaWithDatePattern() {
+        FieldAssembler<Schema> record = SchemaBuilder.record("Main").fields();
+        Schema defaultSchema = record.endRecord();
+        AvroUtils.setIncludeAllFields(defaultSchema, true);
+        AvroUtils.setProperty(defaultSchema, SchemaConstants.TALEND_COLUMN_PATTERN, DATE_PATTERN);
+        return defaultSchema;
+    }
+
     private static final class TestingRuntimeContainer extends DefaultComponentRuntimeContainerImpl {
```
gonchar-ivan: Seems, it can be removed. After return properties refactoring getCurrentComponentId() is not required anymore

components-splunk/src/test/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriterTestIT.java
```diff
@@ -45,7 +49,13 @@
...
 
-    public void testWritingRecords(int recordCount) throws IOException {
+    private final static String DATE_PATTERN = "dd-MM-yyyy";
+
+    public void testWritingRecords(int recordCount, IndexedRecord record) throws IOException {
+        testWritingRecords(recordCount, record, null);
+    }
+
+    public void testWritingRecords(int recordCount, IndexedRecord record, Schema schema) throws IOException {
         DefaultComponentRuntimeContainerImpl container = new TestingRuntimeContainer(COMPONENT_ID);
```
gonchar-ivan: There is code duplication between:
testWritingRecords() 
and
testWritingOneRecordWithWrongToken()

They both setup properties, sink, writeOperation.
So you could move duplication code in separate methods:
setup(create)Properties()
setup(create)Sink(Properties props) // seems you don't need RuntimeContainer 
setup(create)WriteOperation(Sink sink)

As for me it is not good to put "setup" code in test-case method.
In your test-case you only need WriteOperation to create Writer under test
Also you could use @Before annotated method to setup Schema, IndexedRecords etc.

gonchar-ivan:
I am good with this change.

But there are several things, which could be improved in TSplunkEventCollectorWriterTestIT class:
1. Use same naming rules for test cases. There is testWriting1() and testWritingOneRecordWithWrongToken(). Also it could be not easy to guess testWriting1 means test writing one record. It also could mean testWriting(First).
2. Move test arguments setup (Properties, Sink) from test-case method. Test-case method should be as simple and short as possible. It should not contain conditions like if-else.
3. Add description what exactly test-case checks

components-splunk/src/test/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriterTestIT.java
chmyga: Yes, that's totally correct. I don't need a container in a writer also. Will remove it from TSplunkEventCollectorWriter, as it is not used.

components-splunk/src/test/java/org/talend/components/splunk/runtime/TSplunkEventCollectorWriterTestIT.java
chmyga: Properties are different for two of those methods, different values are set to some of their fields, so they cannot be moved to @Before annotated method. Also, I don't see any need in schema creation for every test. I need to create dynamic schema only once for one test.
As for me, "setup" code is the code producing the incoming data, which should be used to produce some expected outgoing data. Setup of this data should be as straightforward as possible, and it should be done for every test separately as every test has it's own set of incoming data.

build-talend:

PR #203, opened 2016-07-15
TDI-36857: Fix salesforce bulk api 10k limitation issue

jzhao-talend:
https://jira.talendforge.org/browse/TDI-36857
Note: for bulk query and bulkExec


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/139/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/139/)

components-salesforce/src/test/java/org/talend/components/salesforce/runtime/SalesforceBulkExecReaderTestIT.java
```diff
@@ -47,16 +48,25 @@
  */
 public class SalesforceBulkExecReaderTestIT extends SalesforceTestBase {
 
+    @Test
+    @Ignore("because of 5M data storage limitation")
```
jzhao-talend: https://resources.docs.salesforce.com/202/latest/en-us/sfdc/pdf/salesforce_app_limits_cheatsheet.pdf
(Page 23)
Because of our salesforce account is developer type. 
So the large number of data load would bread the build.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
```diff
@@ -44,7 +44,7 @@
 
     private JobInfo job;
 
-    private CSVReader baseFileReader;
+    private com.csvreader.CsvReader baseFileReader;
```
gonchar-ivan: 2 different CsvReader from different libraries are used in this class.
Is it correct?

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
```diff
@@ -574,32 +578,49 @@ public String nextResultId() {
 
     class BulkResultSet {
```
gonchar-ivan: Not related to this PR.
Seems, BulkResultSet and BulkResult classes could be static (also they are not required to nested) 
They are not inner logically (they are not using instance variables)

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
jzhao-talend: This just keep same with old talend salesforce api. It's a long history of the different.
Maybe create a jira to investigate it.

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceBulkRuntime.java
jzhao-talend: yes. this can be improvement.

PR #206, opened 2016-07-20
TDI-36929: fix the jar conflict for salesforce and dataprep in new fr…

wwang-talend:
…amework

https://jira.talendforge.org/browse/TDI-36929

### DIFF
--- a/components-parent/pom.xml
+++ b/components-parent/pom.xml
@@ -95,6 +95,16 @@
 				<groupId>org.springframework.boot</groupId>
 				<artifactId>spring-boot-starter</artifactId>
 				<version>1.2.5.RELEASE</version>
+				<exclusions>
+					<exclusion>
+						<groupId>ch.qos.logback</groupId>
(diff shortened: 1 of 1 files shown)


wwang-talend:
the test case fail because : REQUEST_LIMIT_EXCEEDED , so will rerun it tomorrow.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/168/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/168/)

tsteinborn:
why did we exceed the limit? Salesforce allows up to 1000 calls per api a day. So where are we spending all these API calls?

sgandon:
@wwang-talend , do we still have the pb because since the time of this PR the parent changed a bit ?

wwang-talend:
will check it later.

wwang-talend:
now we can run the junit case in the ide like eclipse, no need to change the pom, so we can close it now.

PR #221, opened 2016-08-18
TCOMP-234: added SetupRuntime and SetupComponent

chmyga:
Extracted initialize and validate methods to Runtime interface.
Added SetupRuntime and SetupComponent interfaces.
Extended SourceOrSink interface from Runtime interface.
Added setup method to SetupRuntime without any arguments, because they will be passed to initialize method of parent Runtime interface.


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/275/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/275/)

components-api/src/main/java/org/talend/components/api/component/runtime/Runtime.java
```diff
@@ -0,0 +1,37 @@
...
+import org.talend.components.api.container.RuntimeContainer;
+import org.talend.components.api.properties.ComponentProperties;
+import org.talend.daikon.properties.ValidationResult;
+
+/**
+ * Base interface for defining properties and methods common to Runtime interfaces, like {@link SourceOrSink} or
+ * {@link SetupRuntime}.
+ */
+public interface Runtime {
+
```
francisu: I think the name "Runtime" it too easily confused with the java.lang.runtime. Can we pick a different name?

components-api/src/main/java/org/talend/components/api/component/runtime/SetupRuntime.java
```diff
@@ -0,0 +1,26 @@
...
+ * Basic interface for defining properties and common methods required for Runtimes of components without connectors, or
+ * components which need some pre-processing before performing the actual job.
+ */
+public interface SetupRuntime extends Runtime {
+
+    /**
+     * Perform processes required before main Component's processes or in case a Component doesn't have any connections.
+     */
+    void setup();
+
```
francisu: I don't understand why we need this interface. I think Runtime.validate() could be used for the same purpose, as that's what it's used for with the I/O components, so it would make it more uniform. It also gives you the ValidationResult which can nicely convey a failure.

components-api/src/main/java/org/talend/components/api/component/SetupComponent.java
```diff
@@ -0,0 +1,28 @@
...
+ * Definition of a component which has Setup Runtime. It could be a component which doesn't have any input or output
+ * connectors, or a component which requires some pre-processing.
+ */
+public interface SetupComponent {
+
+    /**
+     * Return the {@link SetupRuntime}.
+     */
+    SetupRuntime getSetupRuntime();
+
```
francisu: Note sure about this interface. Whatever the current mechanism is to get the SourceOrSink should just be generalized to return the "Runtime" interface. No need for this extra method I think.

components-api/src/main/java/org/talend/components/api/component/runtime/Runtime.java
chmyga: Sure, the name is not  that critical. I would be very thankful If you give any suggestions.

components-api/src/main/java/org/talend/components/api/component/runtime/SetupRuntime.java
chmyga: Doing some real work in validate() method is not the best idea, I think. validate should, actually, validate something, make sure that everything is correct. The name of the validate() method doesn't correspond to the method which does some real work.

components-api/src/main/java/org/talend/components/api/component/SetupComponent.java
chmyga: This will be implemented by *ComponentDefinition classes, not SourceOrSink. Also, we assumed that components could be Input, Output, Endpoint and Setup at the same time. So getRuntime() method won't be the best way to call the methods returning runtimes. But it's all should be discussed for sure. I think that Sebastien could give some of his thoughts about this point also.

components-api/src/main/java/org/talend/components/api/component/runtime/Runtime.java
francisu: How about ComponentRuntime?

components-api/src/main/java/org/talend/components/api/component/runtime/SetupRuntime.java
francisu: Sorry, I meant initialize() should handle it, not validate().

components-api/src/main/java/org/talend/components/api/component/SetupComponent.java
francisu: Let's add getRuntime() in ComponentDefinition and have that return a ComponentRuntime object. I think that the InputComponentDefinition can still override that interface and not have to change. Seb is working on a related change that requires a modification to a top-level getRuntime() in ComponentDefinition (I think he needs to take the properties). So we should combine those ideas.

components-api/src/main/java/org/talend/components/api/component/runtime/SetupRuntime.java
chmyga: As far as I understand, initialize() method is used just to initialize some properties and data in the runtime. setup() method is designed to do some work using some objects, which are available after validate() method call. For example a database connection. It is established in validate() method, and then can be used in runtime for other pusrposes. Such purposes could be:
-creating a table in the database;
-cleaning data in database;
-creating some special records in the table;
-downloading/uploading some files to remote servers.
It shouldn't be done before we validate the connection. And also shouldn't be done during validation. it is something that should be processed separately after we initialize() and validate() the runtime.

components-api/src/main/java/org/talend/components/api/component/SetupComponent.java
chmyga: During the discussion with Sebastien, he suggested we could have different method names, so that the Component could return different runtimes at the same time. Also, as I understand, it should be possible to get and run the setup runtime before all the other component tasks are performed. It could be used for preprocessing some data. For example, creating some table, or deleting some data from it. It should be run before the "main" runtime is processed.

components-api/src/main/java/org/talend/components/api/component/runtime/Runtime.java
sgandon: fine by me.

components-api/src/main/java/org/talend/components/api/component/runtime/SetupRuntime.java
sgandon: I agree that this method has a different semantic thatn initialise that is supposed to store the Properties for the runtime and the validate is supposed to validate the component before the process and may even stop the run of the job whereas the *setup* mean it the job is already running and this method is called after the start of the job and everything has been validated.
So we should keep this.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/294/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/294/)

components-api/src/main/java/org/talend/components/api/component/SetupComponent.java
sgandon: As mentioned in the [Jira](https://jira.talendforge.org/browse/TCOMP-234?focusedCommentId=433550&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-433550) SourceOrSink provide methods for getting the schema and schema names so it can not be a generalised *Runtime* interface.
@francisu  I also talked with you of the limitation of the same method name *getRuntime()* returning different object type that was a problem for implementing multiple kind of components into one implem. And you are right we could have a *getRuntime()* in ComponentDefinition, but this brings another question that we need to discuss. if you have a single method that may return different runtime (one source or one sink for instance) how and who will make the choice, how to know the required runtime is for the input comp or the output component.
I would suggest that we keep Dmytro suggestion just for this PR and talk about this on skype to resolve this issue which is a separate matter.

components-api/src/main/java/org/talend/components/api/component/runtime/SetupRuntime.java
francisu: I'm still not in agreement here. The distinction between the 3 methods is still muddy to me. The distinction between initialize() and validate() is clear (as originally defined in GDF), initialize is something entirely local, and validate() will have the component engage with the thing that it's working with (and make sure everything is OK). This is called after the job is started. I don't see a clear distinction between what would go in validate() as opposed to setup(). All of the examples that Dmytro mentioned above would nicely belong in validate() in my estimation; and if any of those things failed, it can report a clear error that the component failed to initialize because of some environmental problem. The connection could be validated as the first thing here. 

@sgandon if you look at this from a DI perspective, the validate() and setup() methods would be called at the same time (in the DI begin method). There is nothing that would happen between the two. 

A legitimate reason for a 3rd method (setup) is that something must happen between validate() and setup(), but that's not the case. I think the GDF people thought though this well, including the names of the methods, and we should just go with their model. Adding this additional method is unnecessary and confusing.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/319/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/319/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/322/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/322/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/325/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/325/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/326/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/326/)

RyanSkraba:
Hello (not the reviewer, just interested)!  This looks like it accurately reflects our long discussion -- in any case, it looks good to me.

A minor objection: I would have preferred to see a `IllegalSpecificationException` or something other than `ValidationResult`.  Exceptions are made for error handling... and "validate" is ruined for us as a verb because of the weird Beam semantics on its validate() method!

But in the end, it doesn't matter _that_ much and I'm fine either way.

gonchar-ivan:
It looks good to me.

components-api/src/main/java/org/talend/components/api/component/runtime/ComponentRuntime.java
```diff
@@ -0,0 +1,37 @@
...
+ * {@link PreProcessRuntime}.
+ */
+public interface ComponentRuntime {
+
+    /**
+     * Gets the ComponentProperties associated with this runtime class in order to configure it. This class also checks
+     * that all required properties are properly set and coherent. No integrity check must be done like checking a
+     * remote connection or checking a file exists here. This method shall be called before the job or flow is executed
+     * and shall prevent execution if the ValidationResult has an ERROR status.
+     *
```
francisu: You should also make it clear here that this method is expected to be called in a distributed environment before the runtime is serialized and sent to other nodes for execution.

components-api/src/main/java/org/talend/components/api/component/runtime/SourceOrSink.java
```diff
@@ -15,42 +15,30 @@
...
- * This allows the {@code SourceOrSink} instance created in this "main program" to be sent (in serialized form) to remote worker
- * machines and reconstituted for each batch of elements being processed. A {@code SourceOrSink} can have instance variable state,
- * and non-transient instance variable state will be serialized in the main program and then deserialized on remote worker
- * machines.
+ * A {@code SourceOrSink} may be passed between partititions for distributed operation and therefore must be
+ * {@code Serializable}. This allows the {@code SourceOrSink} instance created in this "main program" to be sent (in
+ * serialized form) to remote worker machines and reconstituted for each batch of elements being processed. A
+ * {@code SourceOrSink} can have instance variable state, and non-transient instance variable state will be serialized
+ * in the main program and then deserialized on remote worker machines.
  *
```
francisu: Actually these comments should be moved to ComponentRuntime instead of here. This clearly states what's going on and now that's at the level of ComponentRuntime.

components-api/src/main/java/org/talend/components/api/component/runtime/SourceOrSink.java
```diff
@@ -61,4 +49,9 @@
      */
     Schema getEndpointSchema(RuntimeContainer container, String schemaName) throws IOException;
 
+    /**
+     * Checks that this source or sink is valid, before it can be used. This will typically make a connection and return
+     * the results of the connection establishment.
+     */
+    ValidationResult validate(RuntimeContainer container);
```
francisu: This will be called in the same process where the runtime will actually be executed.

components-api/src/main/java/org/talend/components/api/component/runtime/PreProcessRuntime.java
```diff
@@ -0,0 +1,28 @@
...
+ * components which need some pre-processing before performing the actual job.
+ */
+public interface PreProcessRuntime extends ComponentRuntime {
+
+    /**
+     * execute some code before the job or flow is started
+     * 
+     * @throw an RuntimeException if anything fails during the run of this method.
+     */
+    void runPreProcess();
```
francisu: The semantics of this is not clear, especially in a distributed environment. From the previous conversations it seems that this is guaranteed (how?) to be called only once for each entire system to do one-time actions. But is it to be called in the process where this runtime instance will eventually execute? Is this intended for components where no further runtime activity is expected? This all needs to be made clear in the javadoc for the method.

francisu:
@RyanSkraba actually on reflection, I think that ValidationResult is an unfortunate name for that class, as it is really only used to contain error information (beyond just an exception). I don't have a better name though. Maybe something like StatusInformation. Changing the name of that class will hit a lot of things (the Studio code depends on it), so I'm not sure it's worth it at this point, and it certainly should not be considered as part of this PR.

francisu:
@RyanSkraba actually on reflection, I think that ValidationResult is an unfortunate name for that class, as it is really only used to contain error information (beyond just an exception). I don't have a better name though. Maybe something like StatusInformation. Changing the name of that class will hit a lot of things (the Studio code depends on it), so I'm not sure it's worth it at this point, and it certainly should not be considered as part of this PR.

francisu:
@chmyga I'm sorry this is taking so long, I know it must be frustrating. But it really must be right as this is something we are going to live with for a long time. I'm going to be on European time for the next couple of weeks, so we should be able to finish this in the next couple of days, as we can talk real time if necessary, and I will try and be responsive.

components-api/src/main/java/org/talend/components/api/component/runtime/SourceOrSink.java
chmyga: I'm not sure we should move them, as ComponentRuntime is not Serializable and it might not be transferred between nodes in distributed environment(PreProcessRuntime is an example of driver-node only runtime).

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/363/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/363/)

components-api/src/main/java/org/talend/components/api/component/runtime/SourceOrSink.java
francisu: OK, so PreProcessRuntime solves the problem of TCOMP-235, having code that's executed once before the execution of the component starts on the driver node. I think the "PreProcessRuntime" name could be changed to describe that better. And from what I understand, this could be used for any distributed component which would also have a regular Source/Sink. I would prefer the interface be called something like ComponentDriverInitialization, and the method be called initializeAtDriver() to make it very clear what's going on. The flow of execution and the reason this is not serializable should be clearly described in the Javadoc.

components-api/src/main/java/org/talend/components/api/component/runtime/ComponentRuntime.java
```diff
@@ -0,0 +1,41 @@
...
+     * This method is expected to be called in a distributed environment on driver node before the runtime is serialized
+     * and sent to worker nodes for execution.
+     * </p>
+     * 
+     * @param container container instance to get access or share data with the running container.
+     * @param properties used to configure this runtime class.
+     * @return ValidationResult status of the validation, if the status is ERROR then the flow or job shall be stopped.
+     */
+    ValidationResult initialize(RuntimeContainer container, ComponentProperties properties);
+
```
francisu: This should be serializable to satisfy the requirements of TCOMP-234, which allows a distributed component without connectors. It has to be serializable because the properties are to be moved to the cluster nodes. This should not be a conflict with PreProcessRuntime, which itself does not need to be serializable.

Thanks for the improvement of the javadoc, it's very clear now!

components-api/src/main/java/org/talend/components/api/component/runtime/PreProcessRuntime.java
```diff
@@ -0,0 +1,34 @@
...
+     * runtime process. In case of components with connectors, it will be called before serializing and sending runtimes
+     * to worker nodes.
+     * <p>
+     * In case of components without connectors, no other work should be performed for such runtime, only this method
+     * will be called after runtime initialization.
+     * </p>
+     * 
+     * @throw an RuntimeException if anything fails during the run of this method.
+     */
+    void runPreProcess();
```
francisu: Thanks for the javadoc changes, I'm happy with them. As stated below, I would like the interface and method names to change.

components-api/src/main/java/org/talend/components/api/component/runtime/SourceOrSink.java
francisu: This should be moved to ComponentRuntime, as that must be serializable (see previous comment).

chmyga: Maybe we could name the method in some other way, like "runAtDriver()"? In my opinion, developers would think that only some initialization process should be done in method, which includes word"initialize" in its name. But this method would include any different process for components without connectors, for example. So I think the name should be more common than "initialize...". But that is only my opinion. It can always be discussed.

components-api/src/main/java/org/talend/components/api/component/runtime/ComponentRuntime.java
chmyga: PreProcessRuntime should be used for requirements described in TCOMP-234. And I just don't know how will the components without connectors be run on cluster nodes. They aren't Input or Output components, so there's no data passed through them. Maybe I am missing something or we missed something during our discussions?

components-api/src/main/java/org/talend/components/api/component/runtime/SourceOrSink.java
francisu: +1 for runAtDriver()

components-api/src/main/java/org/talend/components/api/component/runtime/ComponentRuntime.java
francisu: A configuration component for example is distributed but not a connector. It would not use PreProcessRuntime, but also have no connections. It would communicate with other components in the worker node using the methods in RuntimeContainer.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/379/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/379/)

components-api/src/main/java/org/talend/components/api/component/runtime/SourceOrSink.java
```diff
@@ -15,42 +14,23 @@
...
-     * Initialize based on the specified properties. This will typically store the {@link ComponentProperties} in the
-     * object.
-     */
-    void initialize(RuntimeContainer container, ComponentProperties properties);
-
-    /**
-     * Checks that this source or sink is valid, before it can be used. This will typically make a connection and return
-     * the results of the connection establishment.
-     */
-    ValidationResult validate(RuntimeContainer container);
```
sgandon: Why did you remove the validate method here, as we need it for beam and use it today to do the integrity checks ?

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/395/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/395/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/396/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/396/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/397/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/397/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/403/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/403/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/404/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/404/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/405/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/405/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/408/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/408/)

PR #227, opened 2016-08-29
TCOMP-346 Implement a single point of registration for components.

RyanSkraba:
Adds DatastoreDefinition and DatasetDefinition interfaces (currently empty)
that will be used to group functions and properties for specific component
families for the Data Streams project.

Adds a ComponentFamilyDefinition that groups all of the Datastores, Datasets,
Components and Wizards for a single family.

Adds a ComponentInstaller interface that provides the single point of
registration for datastores, datasets, components, wizards and families.

Remove all @Component annotations from all classes except ComponentInstallers.

Adapted OSGI and Spring implementations to refer to the ComponentInstaller
to find all definitions.


components-api-archetypes/src/main/resources/archetype-resources/src/test/java/__componentName__Test.java
```diff
@@ -23,9 +23,9 @@
 import org.talend.components.api.component.runtime.Source;
 import org.talend.components.api.exception.error.ComponentsErrorCode;
 import org.talend.components.api.service.ComponentService;
+import org.talend.components.api.service.internal.ComponentRegistry;
 import org.talend.components.api.service.internal.ComponentServiceImpl;
```
RyanSkraba: ?? Should these be moved out of the "internal" package?

components-api/src/main/java/org/talend/components/api/Constants.java
```diff
@@ -13,8 +13,6 @@
 package org.talend.components.api;
 
 public class Constants {
-
-    public static final String COMPONENT_BEAN_PREFIX = "components#";
-
-    public static final String COMPONENT_WIZARD_BEAN_PREFIX = "components#wizards#";
+    /** This is probably not even necessary if there is only one object to collect. */
+    public static final String COMPONENT_INSTALLER_PREFIX = "installer#";
```
RyanSkraba: ?? Should I get rid of this class entirely, now that there's only one instance being installed (and we never need to refer to it by name, just the definitions that it installs...).

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/329/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/329/)

PR #228, opened 2016-08-29
TCOMP-246 Implement a single point of registration for components.

RyanSkraba:
Adds DatastoreDefinition and DatasetDefinition interfaces (currently empty)
that will be used to group functions and properties for specific component
families for the Data Streams project.

Adds a ComponentFamilyDefinition that groups all of the Datastores, Datasets,
Components and Wizards for a single family.

Adds a ComponentInstaller interface that provides the single point of
registration for datastores, datasets, components, wizards and families.

Remove all @Component annotations from all classes except ComponentInstallers.

Adapted OSGI and Spring implementations to refer to the ComponentInstaller
to find all definitions.


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/331/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/331/)

RyanSkraba:
Hello!  I have some outstanding questions to keep in mind while reviewing:

1. `Component` versus `ComponentFramework`: 

There's some classes like `ComponentDefinition` that exclusively deal with literal components on the screen (an action or processing on data), and other classes like `ComponentInstaller` that is really an installer of *Components* and *Component-related things* (like datastores, wizards, and any future definitions we make up).

Is this OK, or should we rename the latter somehow to make this more clear...  (`ComponentInstaller` to  `ComponentFrameworkInstaller` or `FrameworkInstaller` or `TCompInstaller` or `???`).

In the same line: `ComponentRegistry`, `ComponentService`, `ComponentFamilyDefinition`...  all deal with components and component-related things.

2. There are some public classes in `org.talend.components.api.service.*internal*` like ComponentRegistry -- mostly used by test classes.  Is that an issue?

3. I chose to have the interfaces `ComponentInstaller` separate from `ComponentFamilyDefinition` because we can expect some definitions to be installed outside a family (a new file format definition will be used by the existing File Input family components, for example).  What do you think?

components-api-archetypes/src/main/resources/archetype-resources/src/main/java/__componentName__FamilyDefinition.java
```diff
@@ -0,0 +1,89 @@
...
+import org.talend.components.api.datastore.DatastoreDefinition;
+import org.talend.components.api.wizard.ComponentWizardDefinition;
+import org.talend.daikon.i18n.I18nMessages;
+
+import aQute.bnd.annotation.component.Component;
+
+/**
+ * Install all of the definitions provided for the ${componentName} family of components.
+ */
+@Component(name = ${componentName}FamilyDefinition.NAME, provide = ComponentInstaller.class)
```
sgandon: I really believe you should be using some kind of prefix to identify the type of bean registered either for Spring or OSGI service, it is a good practice in order to easily identify the desired beans/services among the myriad of others.
Avoid using # as I happen to learn it not recommended for OSGI.

components-api/src/main/java/org/talend/components/api/Constants.java
```diff
@@ -13,8 +13,6 @@
 package org.talend.components.api;
 
 public class Constants {
-
-    public static final String COMPONENT_BEAN_PREFIX = "components#";
-
-    public static final String COMPONENT_WIZARD_BEAN_PREFIX = "components#wizards#";
+    /** This is probably not even necessary if there is only one object to collect. */
```
sgandon: I think it is required because this service may work with other services in the same container (Spring or OSGI) and this is much easier to identify them with the prefix to tell what kind of beans/services they are and to filter them when necessary.

components-api/src/main/java/org/talend/components/api/ComponentFamilyDefinition.java
```diff
@@ -0,0 +1,34 @@
...
+import org.talend.components.api.datastore.DatastoreDefinition;
+import org.talend.components.api.wizard.ComponentWizardDefinition;
+import org.talend.daikon.NamedThing;
+
+/**
+ * Placeholder for a Component family catalog item.
+ */
+
+public interface ComponentFamilyDefinition extends NamedThing {
+
```
sgandon: I don't really see the need for this interface. The *ComponentFrameworkContext* provide a way to register the different kind of objects handled by the registry and I think it is enough.
If you ask all component family to implement this, this interface will be very hard to maintain and adding new object type will force all components to update even those that will never provide any new object type.

components-api/src/main/java/org/talend/components/api/ComponentInstaller.java
```diff
@@ -0,0 +1,54 @@
...
+     * @param ctx A context object with the available registration methods.
+     */
+    void install(ComponentFrameworkContext ctx);
+
+    /**
+     * The actions that a {@link ComponentInstaller} can take when registering itself to the component framework.
+     */
+    public interface ComponentFrameworkContext {
+
+        void register(DatastoreDefinition datastore);
```
sgandon: following my comment above about me seeing the *ComponentFamilyDefinition* not very necessary, I would suggest those methods take a List<DatastoreDefinition> (and so forth for the other methods) instead of a single object for registration.

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentServiceImpl.java
```diff
@@ -376,7 +365,8 @@ String parseMvnUri(String dependencyString) {
             depsStrings.add("mvn:" + dep.getArtifact().getGroupId() + "/" + dep.getArtifact().getArtifactId() + "/" //$NON-NLS-1$//$NON-NLS-2$ //$NON-NLS-3$
                     + dep.getArtifact().getVersion() + "/"
                     + (dep.getArtifact().getExtension().equals("") ? "" : dep.getArtifact().getExtension())
-                    + (dep.getArtifact().getClassifier().equals("") ? "" : ("/" + dep.getArtifact().getClassifier())));
+                    + (dep.getArtifact().getClassifier().equals("") ? ""
```
sgandon: strange formatting ;)

components-api-archetypes/src/main/resources/archetype-resources/src/main/java/__componentName__FamilyDefinition.java
RyanSkraba: OK, will fix.

components-api/src/main/java/org/talend/components/api/ComponentFamilyDefinition.java
RyanSkraba: This currently doesn't add anything to the "central point of registration" -- this is my interpretation of Francis' request ["have a class for a component family...where the component categorization information can go"](https://jira.talendforge.org/browse/TCOMP-246?focusedCommentId=434999&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-434999).

Right now, this just defines a set of Datastore/Dataset/Component+Wizard that are grouped together, without any additional functions or information for the family (like label/tag/category)...

It's optional and separate from the ComponentInstaller for this reason -- I added the `register(ComponentFamilyDefinition def)` to also register its nested defs as a helper and don't mind removing it.

I don't mind removing it *entirely* and waiting to see if it's really necessary -- if we ever separate labels/tags/families from ComponentDefinitions in the future!

RyanSkraba: Maybe a base class that provides empty lists of nested definitions?  Like `AbstractComponentDefinition` for future proofing new definitions in the future?

(This is where Java 8 default interface implementations would be nice!)

components-api/src/main/java/org/talend/components/api/ComponentInstaller.java
RyanSkraba: Sounds good to me!

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentServiceImpl.java
RyanSkraba: :/ Intellij and the eclipse formatter is usually pretty good... I'll fix this.

components-api/src/main/java/org/talend/components/api/ComponentInstaller.java
RyanSkraba: :P :P If I use List<?> or Iterable<?>, I can't use method overloading to use the same `register` method for each definition.  I'll change these to `registerXxxxDefinition()` ... more explicit is better anyway.

RyanSkraba:
Updated to address sgandon's comments.  I left the ComponentFamilyDefinition in for discussion, and added an abstract implementation.  Let me know if you prefer that I remove it entirely for this PR.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/347/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/347/)

components-api/src/main/java/org/talend/components/api/ComponentFamilyDefinition.java
francisu: I agree with @sgandon I think the ComponentFrameworkContext stuff is fine for this. You can lose this interface.

components-api/src/main/java/org/talend/components/api/ComponentInstaller.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * The actions that a {@link ComponentInstaller} can take when registering itself to the component framework.
+     */
+    public interface ComponentFrameworkContext {
+
+        void registerDatastoreDefinition(Iterable<DatastoreDefinition> defs);
+
+        void registerDatasetDefinition(Iterable<DatasetDefinition> defs);
+
```
francisu: I don't understand datastore and dataset at this level. Both of these seem to be types of ComponentProperties objects and I don't think this should be wired into the Components API. I would support something that can be used to tag COmponentProperties (it could even be an interface) with something and for the constant values of Datastore and Dataset to be in org.talend.components.common, but I don't think these concepts are part of the Components API. The Dataset properties needs to be a generalization of the current SalesforceModule properties and corresponding table properties in the other components. This is the subject of https://jira.talendforge.org/browse/TCOMP-162.

components-api/src/main/java/org/talend/components/api/AbstractComponentFamilyDefinition.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public String getName() {
+        return familyName;
+    }
+
+    @Override
+    protected String getI18nPrefix() {
+        return "family."; //$NON-NLS-1$
+    }
+
```
francisu: I do think that the family name and i18n prefix should be defined in one place, rather than being duplicated for each COmponentDefinition. Perhaps this could be provided by the inner interface in the ComponentInstaller? There needs to be flexibility in the family names. For example some families are in multiple places (like Salesforce) And you should be able to ask the component service to tell you about the supported families. We probably want to have other tagging and categorization information about families which should be in one place and not duplicated in each ComponentDefinition.

components-api/src/main/java/org/talend/components/api/ComponentFamilyDefinition.java
RyanSkraba: Sounds good to me!  Getting rid of ComponentFamilyDefinition and AbstractComponentFamilyDefinition.  Will note in the associated JIRA.

components-api/src/main/java/org/talend/components/api/AbstractComponentFamilyDefinition.java
RyanSkraba: Oops -- this in inconsistent with the comment about getting rid of the ComponentFamilyDefinition.  Can you clarify -- do we want a common structure dedicated to describing a component family?

If we do want a common structure:

* I would prefer to keep it separate from the ComponentInstaller interface.

* The contents of this common structure should be implemented after this PR.  I kept it as empty as possible (just the nested definitions that were part of JIRA TCOMP-246).

francisu: I'm fine with doing it later. I do want something though at some point.

components-api/src/main/java/org/talend/components/api/ComponentInstaller.java
ftom: Sorry Francis, Ryan and I don't understand your point above. We thought we already clarified the needs and interest to distinct between these different sets of metadata for a component family.

francisu: @ftom @RyanSkraba If you refer to https://in.talend.com/14683796, there is this text:

In the existing component framework, the notion of Datastore kind of exists via the ComponentService#getTopLevelWizard().

The notion of Dataset kind of exists for a specific component via embedded ComponentProperties (such as SalesforceModuleProperties), but is not generalized.

(which I agree with). In both of these cases, a dataset and datastore can be simply identified as some marker on a COmponentProperties object. And then we can ask for properties that implement a certain interface (which I think we already have in the component service). So I'm resisting the need of separate APIs for this, or having these concepts be in TCOMP (at least in the API -- they can go in components.common). The hierarchical arrangement between a datastore and a dataset is encoded in the way the properties are implemented, and I would be open to extending the TCOMP API to provide a general means of asking of one type of ComponentProperties can contain another, which would allow this relationship to be expressed.

I hope this clarifies things. If not, let's talk soon. I'm in your time zone now.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/371/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/371/)

RyanSkraba:
@ftom @francisu @sgandon 

I pushed an update so this is current with master.

It looks like there's two issues that need to be addressed:

1) **Removing ComponentFamilyDefinition for this PR**:  

If you're going to want it at some point, I might as well leave it here for now as a placeholder to add future functionality (for categories).  It's useful immediately as the container for related definitions.

I think there's a place for both `ComponentFamilyDefinition` and `ComponentFrameworkContext`.  I imagine that the former "definition" object will be accessible via a service to get information (the group of related components and wizards, but also labels/tags/categories in the future).   The "context" object is only useful during install-time to register the definitions (both grouped in a family and individually).

**For this PR, I propose**: keeping the `ComponentFamilyDefinition` and the best practice for a "normal" component is to provide an implementation very much like the `SalesforceComponentFamilyDefinition`.

2) **Are Datastore and Dataset "definitions" or "interface-tagged properties? **: 

This is the bigger discussion!  I think these have a larger place in the framework than a tag on the properties.

We *could* do the implementation via a tagging interface on a `ComponentProperties`, but a *lot* of the actions that we want to perform on both of these types are very much in parallel with what we do with `ComponentDefinition`, including:

* Getting the list of available Datastore types that TCOMP supports,
* Creating an instance of a Datastore type and persisting it for Data Streams,
* Fetching a runtime based on the values of the instance to make a technology-specific call...

It's weird to me that the Datastore "types" would use inheritance of `ComponentProperties`, while `ComponentDefinition` uses composition.

**For this PR, I propose**: launching a discussion to hash this out -- are you available tomorrow?

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/376/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/376/)

RyanSkraba:
Despite the build-talend message, `mvn clean install` succeeds on this branch!

sgandon:
yes some jira test failed again :(, but now you have conflict with the destination branch, probably because I merged my PR before :)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/386/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/386/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/389/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/389/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/392/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/392/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/414/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/414/)

PR #236, opened 2016-09-09
update the PR review process

sgandon:
I realized that some PR may not be ready for review when created and the current process for reviewing the PR does not cope with this use case.
This is why I suggest to update the process by adding a new label **Need review**.
Please @pteyssier @gonchar-ivan can you have a look and let me know what you think about this.
I will eventually present it to the TAB because it seems to be working well I think.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,3 +30,10 @@ Here is the workflow for pull requests.
   * **Can be merged after rework**, The review is fine but requires some minor (cosmetic) adjustments that do not need to be reviewed again. The PR is then assigned back to the PR's creator for some minor work. Once the work is done the creator can directly Merge and Squash it without the need of any review. If the creator does not have the right to do so, it will assign it back to the Reviewer after rework so that it can be merged.
   * **Rework Required**, the review added some comments that need work to be done and another review is required after. The reviewer assigns it to the PR creator.
 7. if **Rework Required** once assigne to the PR creator, he/she has to make the required changes, then assign it back to the initial Reviewer and set the label **Need review** again. Then back to step 5.
+
+Here are the label definitions for this workflow (label name : color code) :
+* Need review                : #fef2c0
+* Can be merged after rework : #c2e0c6
+* Can be merged              : #0e8a16
+* Reviewing                  : #fbca04
+* Rework required            : #1d76db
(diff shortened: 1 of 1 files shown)


gonchar-ivan:
For now, there are 2 groups of labels:
1. build status: failed and passed
2. review status: Reviewing, Can be merged and all other.

For now, PR need review, when there is no label from "review status" label group.

After this change:
No label will mean "Do not review right now. I want to do something else".
"Need review" label will mean: "Please, review it."

I am ok with this change.
However maybe "On hold" label can be used by developers to mark "Do not review right now".
Also I wonder what are reasons to create PR, if developer doesn't need review right now?

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/432/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/432/)

RyanSkraba:
Hello!  I'm in this situation right now -- I've got an initial implementation that some others *might* be interested to look at, but it is obviously not finished and doesn't require a full review (yet!).

Instead of "On hold", I just stuck it at "Rework required"... as if I started after a failed code review.

The reason I'd stick it as a PR right away, rather than a branch, is for (a) mergeability  and buildability and (b) a better feedback interface for comments!

What do you think?

gonchar-ivan:
Ryan, thanks for good example.

Also PR shows diff in a convenient format.

sgandon:
So what you you think @gonchar-ivan and @RyanSkraba  on using the **Rework Required** instead of a new **onHold** ? I like the idea of using this since it indicates some work in progress.

RyanSkraba:
I think more projects use a tag like "On Hold" or "Work in progress"  (Data Streams uses Work in progress).  I like starting at Rework Required because it follows the process and "review loop" that you've defined already - It just starts at a different place.

sgandon:
or may be we find a suitable name that matches both use cases such as **Work required** ?

gonchar-ivan:
As for me all options are good:
1. Need review
2. Work required
3. Reword required

I have no criteria to prefer one of them between others

sgandon:
but they are not the same, **need review** requires reviewer work and **Re/work required** requires work from the developer.
But I think I got your point.

gonchar-ivan:
Yes. I mean we have several options to cover use-case, when developer doesn't want review right now.
There are:
1. Add "Need review" label.
Then "No label" means nothing to do with this PR
"Neew review" means "Please, review it"

2. Using "Rework required". In this case:
No label means "Please, review it"
"Rework required" could be set by both Developer or Reviewer. It means Developer should do something.

3. The same as 2, but rename label

sgandon:
A new proposal has been adopted : https://github.com/Talend/tools/blob/master/tools-root-github/CONTRIBUTING.md
So I'll update the PR accordingly.

PR #237, opened 2016-09-09
TCOMP-253 Add RuntimableDefinition as an extension for business objects.

RyanSkraba:


components-api/src/main/java/org/talend/components/api/RuntimableDefinition.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * This can be used to provide "extension" business objects in the component framework, using the same access pattern as
+ * {@link org.talend.components.api.component.ComponentDefinition} instances. This will usually be done by providing a
+ * common sub-interface of this class that a set of component families know how to interpret.
+ *
+ * @param <PropT> The type of properties that the specific definition works with.
+ * @param <RuntimeInfoContextT> Creating a runtime object for this class may require additional information than just
+ * the properties.
+ */
+public interface RuntimableDefinition<PropT extends Properties, RuntimeInfoContextT> {
```
sgandon: I think this should also extend NamedThing to have name and a display name to show to the user when showing the possible datastore or dataset available.
I also think you should push up the **String getPngImagePath(ComponentImageType imageType);** from ComponentDefiniton

components-common/src/main/java/org/talend/components/common/dataset/DatasetProperties.java
```diff
@@ -0,0 +1,9 @@
+package org.talend.components.common.dataset;
+
+import org.talend.daikon.properties.Properties;
+
+/**
+ * Placeholder for DatasetProperties.
+ */
+public interface DatasetProperties extends Properties {
```
sgandon: Do you really need to have a specific subclass of properties ? especially if it has no specialization on top of Properties.

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentRegistry.java
```diff
@@ -30,6 +29,8 @@
  */
 public class ComponentRegistry implements ComponentInstaller.ComponentFrameworkContext {
 
+    private List<RuntimableDefinition<?, ?>> extensions;
+
     private Map<String, ComponentDefinition> components;
```
sgandon: Do you need to separates extensions from ComponentDefinition and ComponetWizardDefitnion ? I guess all could leave in on single list.

francisu: I agree with this, all of the definitions should be one thing and then you can tell from the class which kind it is.

components-common/src/main/java/org/talend/components/common/dataset/DatasetProperties.java
francisu: This will be the basis for the https://jira.talendforge.org/browse/TCOMP-162 solution, right? Please make sure that solving TCOMP-162 is part of your overall proposals for dataset. We want to convert the existing components to use this mechanism.

components-api/src/main/java/org/talend/components/api/RuntimableDefinition.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * @return An empty instance of the properties for an instance of this definition.
+     */
+    PropT createProperties();
+
```
francisu: I think we also should move the "supportsProperties" method up from ComponentDefinition to here. Making it Properties of course.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/441/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/441/)

components-api/src/main/java/org/talend/components/api/RuntimableDefinition.java
RyanSkraba: If at all possible -- I'd rather not!

My justification is that you should never be looking at a `RuntimableDefinition` without knowing whether it is a `DatasetDefinition`, `DatastoreDefinition` or `OtherDefinition` beforehand...!

If you want those methods, they should be in the Datastore/Dataset/OtherDefinition, not here.

(ComponentImageType is also `ComponentDefinition`-specific).

components-common/src/main/java/org/talend/components/common/dataset/DatasetProperties.java
RyanSkraba: The required common configuration will go here (returned via the `createProperties` method in the `DatasetDefinition`).

The required common behaviour will go in the object returned via the `getRuntimeInfo` method in the `DatasetDefinition`.

Keep in mind that I don't want to solve TCOMP-162 in this PR :)  Just put some more pieces in place for these problems to be solved!

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentRegistry.java
RyanSkraba: Unfortunately, `ComponentWizardDefinition` doesn't have `Properties` or `RuntimeInfo`!  It can't be in the same list.

I'll take a look at `ComponentDefinitions` -- it's currently in the `List` as well, and the `Map` is just for the specialized endpoint that looks up by name.

components-api/src/main/java/org/talend/components/api/RuntimableDefinition.java
RyanSkraba: See above -- I'm not keen to enforce a contract on anybody here if there's a better place to put it.

components-api-service/src/main/java/org/talend/components/api/service/internal/spring/ComponentServiceSpring.java
```diff
@@ -20,6 +20,7 @@
 
 import javax.servlet.http.HttpServletResponse;
 
+import avro.shaded.com.google.common.collect.Lists;
```
sgandon: I am not sur is is desirable to have an avro dependency here.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/443/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/443/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/444/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/444/)

components-api/src/main/java/org/talend/components/api/RuntimableDefinition.java
francisu: I think we want to support searches on any definitions that supports any type of properties. That's the point of supportsProperties(). This is why I think it's the correct place.

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentRegistry.java
```diff
@@ -30,6 +31,16 @@
...
+     * All of the {@link RuntimableDefinition}s that have been added to the framework, including
+     * {@link ComponentDefinition}s.
+     */
+    private List<RuntimableDefinition<?, ?>> definitions;
+
+    /**
+     * All of the {@link ComponentDefinition}s that have been added to the framework, keyed by name. The values in this
+     * map are all in the {@link #definitions} member.
+     */
     private Map<String, ComponentDefinition> components;
```
RyanSkraba: Added comments to clarify ... 

I can get rid of it, but it should be useful support the endpoints specific to ComponentDefinition.

Is this acceptable for now?

components-api/src/main/java/org/talend/components/api/RuntimableDefinition.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * This can be used to provide "extension" business objects in the component framework, using the same access pattern as
+ * {@link org.talend.components.api.component.ComponentDefinition} instances. This will usually be done by providing a
+ * common sub-interface of this class that a set of component families know how to interpret.
+ *
+ * @param <PropT> The type of properties that the specific definition works with.
+ * @param <RuntimeInfoContextT> Creating a runtime object for this class may require additional information than just
+ * the properties.
+ */
+public interface RuntimableDefinition<PropT extends Properties, RuntimeInfoContextT> extends NamedThing {
```
RyanSkraba: Now extends NamedThing!

components-api/src/main/java/org/talend/components/api/RuntimableDefinition.java
RyanSkraba: Would it be acceptable to leave this change for TCOMP-270 then, instead of now?  Or the first feature that will need supportsProperties()?

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentRegistry.java
sgandon: I am not very fond of having 2 containers for the same instances.

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentRegistry.java
```diff
@@ -30,6 +31,16 @@
...
 
     private Map<String, ComponentWizardDefinition> componentWizards;
```
sgandon: 👍

components-api/src/main/java/org/talend/components/api/AbstractComponentFamilyDefinition.java
```diff
@@ -78,12 +87,17 @@ protected String getI18nPrefix() {
     }
 
     @Override
-    public Iterable<ComponentDefinition> getComponents() {
+    public Iterable<? extends ComponentDefinition> getComponents() {
```
sgandon: this seems awkward to still have a difference between ComponentDefinition and Extensions here.

components-api/src/main/java/org/talend/components/api/ComponentFamilyDefinition.java
```diff
@@ -19,10 +19,11 @@
 /**
  * Placeholder for a Component family catalog item.
  */
-
 public interface ComponentFamilyDefinition extends NamedThing {
 
-    Iterable<ComponentDefinition> getComponents();
+    Iterable<? extends ComponentDefinition> getComponents();
```
sgandon: here too, no need to have this just use the *getExtensions* but may be rename it to *getDefintions* or *getRuntimeableDefintions*

sgandon:
oups sorry, I just approved the changes that where not made. sorry just testing the review process.

components-api/src/main/java/org/talend/components/api/AbstractComponentFamilyDefinition.java
RyanSkraba: No problem, will remove.

components-api/src/main/java/org/talend/components/api/ComponentFamilyDefinition.java
RyanSkraba: No problem, will rename to getDefinitions().

components-api/src/main/java/org/talend/components/api/ComponentFamilyDefinition.java
```diff
@@ -19,10 +19,11 @@
...
+
+    Iterable<? extends ComponentWizardDefinition> getComponentWizards();
 
-    Iterable<ComponentWizardDefinition> getComponentWizards();
+    Iterable<? extends RuntimableDefinition<?, ?>> getExtensions();
```
RyanSkraba: Will remove.

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentRegistry.java
RyanSkraba: Will remove

francisu:
@RyanSkraba (sorry for the delay), yes, we can wait on the supportProperties thing until it's needed.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/473/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/473/)

components-api-full-example/src/main/java/org/talend/components/fullexample/datastore/FullExampleDatastoreProperties.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Properties that can be used to configure a FullExampleDatastore.
+ */
+public class FullExampleDatastoreProperties extends PropertiesImpl implements DatastoreProperties {
+
+    public Property<String> host = PropertyFactory.newString("host");
+
+    public Property<Integer> port = PropertyFactory.newInteger("host");
+
```
francisu: Given that we have HostPort as a common set of properties you use that class, or pick something different. In any case you don't want "host" twice here.

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentRegistry.java
```diff
@@ -12,15 +12,16 @@
...
+import java.util.*;
 
 import org.talend.components.api.ComponentFamilyDefinition;
 import org.talend.components.api.ComponentInstaller;
+import org.talend.components.api.RuntimableDefinition;
 import org.talend.components.api.component.ComponentDefinition;
 import org.talend.components.api.wizard.ComponentWizardDefinition;
 
+import com.google.common.collect.Iterables;
+
```
francisu: We are introducing com.google.common.collect dependency? Is this new? I want to be careful about TCOMP dependencies.

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentServiceImpl.java
```diff
@@ -166,13 +178,8 @@ public InputStream getWizardPngImage(String wizardName, WizardImageType imageTyp
...
-        ComponentDefinition componentDefinition = componentRegistry.getComponents().get(componentName);
-        if (componentDefinition != null) {
-            return getImageStream(componentDefinition, componentDefinition.getPngImagePath(imageType));
-        } else {
-            throw new ComponentException(ComponentsApiErrorCode.WRONG_COMPONENT_NAME,
-                    ExceptionContext.build().put("name", componentName)); //$NON-NLS-1$
-        }
+        ComponentDefinition componentDefinition = getComponentDefinition(componentName);
+        return getImageStream(componentDefinition, componentDefinition.getPngImagePath(imageType));
     }
```
francisu: Nice refactoring! And thanks for catching the spelling errors.

components-api/src/main/java/org/talend/components/api/AbstractComponentFamilyDefinition.java
```diff
@@ -43,22 +43,22 @@ public AbstractComponentFamilyDefinition(String familyName, Object... definition
 
         if (definitions == null || definitions.length == 0) {
             // Shortcut if there are no definitions. The subclass can overwrite the implementations instead.
-            this.components = Collections.<ComponentDefinition> emptyList();
-            this.componentWizards = Collections.<ComponentWizardDefinition> emptyList();
+            this.definitions = Collections.emptyList();
+            this.componentWizards = Collections.emptyList();
         } else {
```
francisu: Please remove "this" unless required.

components-api/src/main/java/org/talend/components/api/RuntimableDefinition.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.talend.components.api.component.runtime.RuntimeInfo;
+import org.talend.daikon.NamedThing;
+import org.talend.daikon.properties.Properties;
+
+/**
+ * A base class for definitions of business objects provided by the component framework.
+ *
+ * This can be used to provide "extension" business objects in the component framework, using the same access pattern as
+ * {@link org.talend.components.api.component.ComponentDefinition} instances. This will usually be done by providing a
```
francisu: I think this "extension" terminilogy was for an earlier version of this when you had the notion of extension objects or some thing. Can you adjust this comment to refer to Properties which will not necessarily manifest to be a component, but can be used by components? That will make it more clear.

components-api-service/src/main/java/org/talend/components/api/service/internal/ComponentRegistry.java
RyanSkraba: Hello!  It's unfortunate, because the guava library does *exactly* what we want -- and it's already imported from the swagger dependency...

However, I noticed too late that the swagger dependency is "provides" scope, and not included in the OSGi bundle (I'm not sure about the reasoning behind that, I'm guessing that Swagger is necessary to compile but not to run!).

I'd be just as happy to remove this dependency since it only exists in this one place.

sgandon: We may eventually remove swagger to use javadoc annotations along with *enunciate*

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/485/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/485/)

sgandon:
LGTM
👍

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/508/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/508/)

PR #239, opened 2016-09-13
Feature/6.3/tcomp 230 flow variables definition. Helper method

gonchar-ivan:
Related to https://jira.talendforge.org/browse/TCOMP-230

We already have support of After Return variables (We call them simply Return values/properties).
However Return values are not only After variables. There is another type of Return values - Flow variables. They are often used in Studio components.

This PR adds helper method to set type (availability) of return property. Type could be AFTER or FLOW.

### DIFF
--- a/components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
+++ b/components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
@@ -192,5 +192,26 @@ public abstract class AbstractComponentDefinition extends AbstractTopLevelDefini
         }
         return namedThings;
     }
+    
+    /**
+     * Types (availability) of Return properties
+     * This type means in what point of Job execution Return variable is available.
+     * AFTER variable is available after all records are processed in current flow
+     * FLOW variable is available after each record is processed in current flow
+     */
+    protected static enum ReturnType {
+        AFTER,
+        FLOW
+    }
+    
+    /**
+     * Sets {@link ReturnType} of Return property.
+     * 
+     * @param property Return property, which type to be set
+     * @param type {@link ReturnType}
+     */
+    protected static void setReturnType(Property<?> property, ReturnType type) {
+        property.setTaggedValue("AVAILABILITY", type.toString());
+    }
 
 }


build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/435/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/435/)

francisu:
There is an existing return mechanism in TCOMP (in RuntimeContainer), and this seems to be done in a very separate way. We could implement this by adding a new constant in RuntimeContainer that handles the per-record case. Has this been looked into?

gonchar-ivan:
Thank you for mentioning RuntimeContainer. It brings more ideas.
There is globalMap mechanism in Studio. It is used for sharing some information.
Both after and flow variables (also known as return properties) are just globalMap key-value pairs.
The difference between them only in moment when they are retrieved from component and stored in globalMap.
Some existent Jobs can contain "globalMap.get(returnVariableKey, returnVariableValue)"
So. updated(ported) components should also store values in globalMap in corresponding moments of time.
There is no notion of globalMap in Component Framework, but there is RuntimeContainer (which is actually wrapper around globalMap in Studio generated code)

So, I see 2 options to put values in RuntimeContainer:
1. Inside component. So component (Reader/Writer/etc) should store an instance of RuntimeContainer. 
2. Outside component. Environment/Platform calls getReturn/Flow/After/Values() method and stores retrieved values in RuntimeContainer/globalMap.

Here, I propose 2nd option

francisu:
@gonchar-ivan why not just add a new constant to RuntimeContainer for your flow values? The purpose of the RuntimeContainer is to represent the environment/container in which your component is running?

gonchar-ivan:
So, here it is not done in separate way... or I didn't understand you.

sgandon:
I think those FLow variable should be set using the RuntimeContainer global variable, may be using some king of prefix in the key to identify them and convert them in the javajet to actual flow variable.
And yes this means you should store the container instance in you Reader/Writer if required.

gonchar-ivan:
I distinguish producers and consumers of flow variables. This feature is required for compatibility with DI designer. In DI after/flow variables sharing is performed via globalMap instance. Producers stores data to globalMap, consumers retrieve data from it.

It is good idea to use RuntimeContainer for flow variables. In DI RuntimeContainer is wrapper on globalMap. However, flow variables are produced in Reader/Writer, which implement Serializable. RuntimeContainer is not Serializable. RuntimeContainer represents environment and shouldn't be Serializable and stored in Reader/Writer.

That's why we have added API to retrieve After values. Please, loot at following code:
Writer.close:
https://github.com/Talend/components/blob/master/components-api/src/main/java/org/talend/components/api/component/runtime/Writer.java#L60
WriteOperation.finalize:
https://github.com/Talend/components/blob/master/components-api/src/main/java/org/talend/components/api/component/runtime/WriteOperation.java#L75

Compare TCOMP finalize with Beam's finalize:
https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/Sink.java#L191

Beam's finalize doesn't return any value, but we have added it (seems, the only reason is Return properties).

Also check corresponding codegen:
https://github.com/Talend/tdi-studio-se/blob/master/main/plugins/org.talend.designer.codegen/jet_stub/generic/component_end.javajet#L45-L55

So, for after values: we retrieve them from component (without RuntimeContainer usage).
I propose to go similar way with Flow variables, but retrieve these values per record.

gonchar-ivan:
About adding new constant for storing in RuntimeContainer.
First of all, it is not intended to store in RuntimeContainer (because it is not stored in Reader/Writer)
The second reason, we can store these values directly in globalMap in codegen.


In DI designers flow/after variable key looks like:
"tSomeComponentName_1_VARIABLE_NAME"
"tSomeComponentName_1_VARIABLE_NAME"

"prefix_VARIABLE_NAME" -> ""

It consist of 3 parts:
1. component name
2. Number, which index number of this specific component (it denotes a quantity of this component instances in job)
3. Variable name
1+2 = <%=cid%>
cid variable is generated in codegen.

components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
```diff
@@ -192,5 +192,26 @@ public String toString() {
...
+        FLOW
+    }
+    
+    /**
+     * Sets {@link ReturnType} of Return property.
+     * 
+     * @param property Return property, which type to be set
+     * @param type {@link ReturnType}
+     */
+    protected static void setReturnType(Property<?> property, ReturnType type) {
```
gonchar-ivan: This change adds means to type of Return property (could be After or Flow) for definition reasons.
This type is displayed in Outline tab in Studio. It gives tips for users how to use this Return value

components-api/src/main/java/org/talend/components/api/component/runtime/HasReturnValues.java
```diff
@@ -0,0 +1,48 @@
...
+ * 
+ * Return Variables is special technical data, which could be retrieved from component during runtime:
+ * either during data flow (Flow variables), or after the component has completed (After variables).
+ * These variables store different information like error message, number of record processed, number of failed records etc.
+ * 
+ * Runtime Platform should check whether Component implements this interface and retrieve return values from it.
+ * After variables should be retrieved after all records were processed and component finished its work.
+ * Flow variables should be retrieved after each record processed
+ */
+public interface HasReturnValues {
```
gonchar-ivan: This API is optional in 2 senses:
1. Component may not implement it, if it doesn't provide flow/after variables
2. Environment may not call it, if it doesn't support flow/after variables.

This interface is intended to be implemented by Reader and Writer

components-api/src/main/java/org/talend/components/api/component/runtime/HasReturnValues.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    Map<String, Object> getFlowValues();
+
+    /**
+     * Returns map of after values or null if they are not available
+     *
+     * @return map with current flow values. Flow variable key should be equal to corresponding Return property
+     * name from {@link ComponentDefinition}
+     */
+    Map<String, Object> getReturnValues();
```
gonchar-ivan: This is the same method as Reader.getReturnValues().

I want to unify this method and make it common for Reader/Writer and other Runtime classes

components-api-full-example/src/main/java/org/talend/components/fullexample/FullExampleDefinition.java
```diff
@@ -36,7 +45,7 @@ public FullExampleDefinition() {
 
     @Override
     public Property[] getReturnProperties() {
-        return new Property[0];
+        return new Property[] { RETURN_ERROR_MESSAGE_PROP, RETURN_FLOW_EXAMPLE_PROP};
```
gonchar-ivan: This is an example of flow/after variables definition

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/502/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/502/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/503/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/503/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/52/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/52/)

francisu:
I think this PR is not aligned with what we have previously agreed on handling flow variables.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/182/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/182/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/192/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/192/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/194/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/194/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/195/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/195/)

gonchar-ivan:
It is no longer relevant.
See TCOMP-230 for details

PR #243, opened 2016-09-20
TDI-37456: Need a means to allow for different sets of runtime libraries

wwang-talend:
for a component configuration
https://jira.talendforge.org/browse/TDI-37456


wwang-talend:
the same content with the close one, only rebase on the lastest master

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/480/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/480/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/482/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/482/)

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/JDBCRowSourceOrSink.java
```diff
@@ -0,0 +1,157 @@
...
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.List;
+
+import org.apache.avro.Schema;
+import org.talend.components.api.component.runtime.SourceOrSink;
+import org.talend.components.api.container.RuntimeContainer;
+import org.talend.components.api.exception.ComponentException;
+import org.talend.components.api.properties.ComponentProperties;
+import org.talend.components.jdbc.ComponentConstants;
```
gonchar-ivan: I prefer to store constants in public final class, not in interface
Interface is intended to be implemented by other class. It provides API for clients..
However, it's a matter of taste

components-jdbc/pom.xml
```diff
@@ -34,6 +34,13 @@
         </dependency>
         
         <dependency>
+            <groupId>org.talend.components</groupId>
```
gonchar-ivan: I've found <artifactId>javax.inject</artifactId> dependency in pom.xml

Do you really need this dependency?

gonchar-ivan: the same about <artifactId>json-simple</artifactId>

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/JDBCRowSourceOrSink.java
```diff
@@ -0,0 +1,157 @@
...
+    public List<NamedThing> getSchemaNames(RuntimeContainer runtime) throws IOException {
+        return null;
+    }
+
+    @Override
+    public Schema getEndpointSchema(RuntimeContainer runtime, String tableName) throws IOException {
+        return null;
+    }
+
+    public Connection connect(RuntimeContainer runtime) throws ClassNotFoundException, SQLException {
```
gonchar-ivan: Why it is public? Is it intended to use outside this class?

gonchar-ivan:
I'm not sure it is allowed to establish connection in SourceOrSink.validate() method. Because this method has special goal in Apache Beam framework. 
There was a discussion about it with Seb and Francis. We should clarify, where connection things should be implemented.

Thus, I haven't finished review yet

sgandon:
About establishing connection in Validate, there is no clear status on the beam side so you may keep it in until we have more information.

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/JDBCRowSourceOrSink.java
```diff
@@ -0,0 +1,157 @@
...
+import org.talend.components.jdbc.JDBCConnectionInfoProperties;
+import org.talend.components.jdbc.tjdbcrow.TJDBCRowProperties;
+import org.talend.daikon.NamedThing;
+import org.talend.daikon.properties.ValidationResult;
+
+public class JDBCRowSourceOrSink implements SourceOrSink {
+
+    private static final long serialVersionUID = -1730391293657968628L;
+
+    public TJDBCRowProperties properties;
```
francisu: Properties should be serializable. It's intended to be sent on the wire or stored. It can be directly referred to by the SOurceOrSink.

francisu:
Validate is intended exactly for things like establishing the connection. It's designed to be called to make sure the component is actually usable. If Beam ignores the failed status of validation, that's a separate concern we need to try and fix in that environment.

sgandon:
@francisu Ryan pointed that beam has 2 documents about the validate method that contradict with each other, this is why this is not clear.

components-jdbc/src/main/java/org/talend/components/jdbc/module/DriverTable.java
```diff
@@ -0,0 +1,31 @@
...
+import static org.talend.daikon.properties.property.PropertyFactory.newProperty;
+
+import java.util.List;
+
+import org.apache.commons.lang3.reflect.TypeLiteral;
+import org.talend.components.api.properties.ComponentPropertiesImpl;
+import org.talend.daikon.properties.presentation.Form;
+import org.talend.daikon.properties.property.Property;
+
+// have to implement ComponentProperties, not good
```
wwang-talend: it's about the i18n, if use Properties, meet some issue about 18n.

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/JDBCRowSourceOrSink.java
wwang-talend: yes, as we create the connection in SourceOrSink.validate() method, and use the same connection later in Source.start or  Sink.open method.

wwang-talend: in the common javajet, we always call validate method before start or open method

components-jdbc/pom.xml
wwang-talend: will check it later, not sure will fix it at once as it's not so important. And it's necessary to remove it if that.

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/JDBCRowSourceOrSink.java
wwang-talend: i think they are both ok,  agree with you. But i think it don't affect the code reading and they are both easy to maintain .

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/writer/JDBCRowWriter.java
```diff
@@ -0,0 +1,330 @@
...
+
+        IndexedRecord output = new GenericData.Record(outSchema);
+        for (Schema.Field outField : output.getSchema().getFields()) {
+            Object outValue = null;
+
+            if (propagateQueryResultSet) {
+                String columnName = properties.useColumn.getValue();
+                if (outField.name().equals(columnName)) {
+                    output.put(outField.pos(), resultSet);
+                    continue;
```
chmyga: condition can be changed to something like 
if (propagateQueryResultSet && outField.name().equals(properties.useColumn.getValue()))
Then you can use "else" after this condition instead of using "continue".

components-jdbc/src/main/java/org/talend/components/jdbc/tjdbcrow/TJDBCRowDefinition.java
```diff
@@ -41,19 +45,41 @@ public TJDBCRowDefinition() {
...
+        switch (connectorTopology) {
+        case OUTGOING:
+            return JDBCTemplate.createCommonRuntime(this.getClass().getClassLoader(), properties,
+                    JDBCRowSource.class.getCanonicalName());
+        case INCOMING:
+            return JDBCTemplate.createCommonRuntime(this.getClass().getClassLoader(), properties,
+                    JDBCRowSink.class.getCanonicalName());
+        case INCOMING_AND_OUTGOING:
+            return JDBCTemplate.createCommonRuntime(this.getClass().getClassLoader(), properties,
+                    JDBCRowSink.class.getCanonicalName());
```
chmyga: those two cases can use the same return, something like:
case INCOMING:
case INCOMING_AND_OUTGOING:
    return JDBCTemplate.createCommonRuntime(this.getClass().getClassLoader(), properties,
                    JDBCRowSink.class.getCanonicalName());

components-jdbc/src/test/java/org/talend/components/jdbc/JDBCConnectionTestIT.java
```diff
@@ -51,20 +55,15 @@ public static void init() throws Exception {
             props.load(is);
         }
 
+        // driverPath = props.getProperty("driverPath");
```
chmyga: It's not a good practice to leave the commented out lines like that. Especially in tests. You could leave them in code in case some bugs appear, but you shouldn't do that in tests. Better to just remove them.

components-jdbc/src/test/java/org/talend/components/jdbc/common/DBTestUtils.java
```diff
@@ -622,4 +632,34 @@ public static DataAction randomDataActionExceptDelete() {
         }
     }
 
+    public static Schema createTestSchema4() {
+        FieldAssembler<Schema> builder = SchemaBuilder.builder().record("TEST").fields();
+
+        Schema schema = AvroUtils._string();// TODO : fix it as should be object type
```
chmyga: Is this TODO left for the future or you just forgot about it?

wwang-talend: no object type in AvroUtils now

wwang-talend: for the future

components-jdbc/src/test/java/org/talend/components/jdbc/JDBCConnectionTestIT.java
wwang-talend: i leave them here for the replacement in future, will remove it now.

components-jdbc/src/main/java/org/talend/components/jdbc/tjdbcrow/TJDBCRowDefinition.java
wwang-talend: fix it now

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/writer/JDBCRowWriter.java
wwang-talend: i like "continue" more than "else", it's caused by the javajet.  Agree with the part for using only one "if".

wwang-talend: seems "else" is more clear here, use it now

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/reader/JDBCRowReader.java
```diff
@@ -0,0 +1,210 @@
...
+        IndexedRecord reject = new GenericData.Record(outSchema);
+
+        for (Schema.Field outField : reject.getSchema().getFields()) {
+            Object outValue = null;
+
+            if ("errorCode".equals(outField.name())) {
+                outValue = e.getSQLState();
+            } else if ("errorMessage".equals(outField.name())) {
+                outValue = e.getMessage();
+            }
```
wwang-talend: Agree, the "errorMessage" work for many components, will fix it later.

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/reader/JDBCRowReader.java
```diff
@@ -0,0 +1,210 @@
...
+
+        if (propagateQueryResultSet) {
+            String columnName = properties.useColumn.getValue();
+            for (Schema.Field outField : output.getSchema().getFields()) {
+                if (outField.name().equals(columnName)) {
+                    output.put(outField.pos(), resultSet);
+                }
+            }
+        }
+
```
wwang-talend: 1. Propagating the result set is the unique reason for : the component work as a input component, no meaning for setting value to the other columns.

2. Agree, need a context to consider it, will check it later.

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/reader/JDBCRowReader.java
```diff
@@ -0,0 +1,210 @@
...
+            String sql = properties.sql.getValue();
+            boolean propagateQueryResultSet = properties.propagateQueryResultSet.getValue();
+
+            if (usePreparedStatement) {
+                prepared_statement = conn.prepareStatement(sql);
+
+                JDBCTemplate.setPreparedStatement(prepared_statement, properties.preparedStatementTable);
+
+                if (propagateQueryResultSet) {
+                    resultSet = prepared_statement.executeQuery();
```
wwang-talend: result set object may be used by the next component like tParseRecordSet, so can't close it.

wwang-talend: Agree with that : we should limit the scope for every var, so the field is evil sometimes, but it's good sometimes for code refactor or reading. So i will keep it now and only change the "protected" to "private" for the field.

PR #246, opened 2016-09-22
TCOMP-257 : Use json schema/ui schema/json data instead of the jsonio for ComponentProperties

bchen-talend:


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/3/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/3/)

components-api-service/src/main/java/org/talend/components/api/service/util/JsonGenerator.java
```diff
@@ -0,0 +1,440 @@
...
+        try {
+            aClass = Class.forName(className);
+        } catch (ClassNotFoundException e) {
+            int lastPosition = className.lastIndexOf(".");
+            className = className.substring(0, lastPosition) + "$" + className
+                    .substring(lastPosition + 1);
+            try {
+                aClass = Class.forName(className);
+            } catch (ClassNotFoundException e1) {
+                e1.printStackTrace();
```
sgandon: please use SL4J for login issues, and this will cause an NPE in the called because null will be returned.

components-api-service/src/main/java/org/talend/components/api/service/util/JsonGenerator.java
```diff
@@ -0,0 +1,440 @@
...
+
+    private List<Property> listTProperty(Properties cProperties) {
+        List<Property> propertyList = new ArrayList<>();
+        Field[] allFields = cProperties.getClass().getDeclaredFields();
+        for (Field field : allFields) {
+            if (Property.class.isAssignableFrom(field.getType())) {
+                try {
+                    propertyList.add((Property) field.get(cProperties));
+                } catch (IllegalAccessException e) {
+                    e.printStackTrace();
```
sgandon: no **e.printstackTrace()** please.

components-api-service/src/main/java/org/talend/components/api/service/util/JsonWidget.java
```diff
@@ -0,0 +1,38 @@
...
+public class JsonWidget {
+    private Widget widget;
+    private Form form;
+
+    public JsonWidget(Widget widget, Form form) {
+        this.widget = widget;
+        this.form = form;
+    }
+
+    public int getOrder() {
```
sgandon: please add some documentation to explain what you do.

components-api-service/src/test/java/org/talend/components/api/service/util/TestComponentProperties.java
```diff
@@ -0,0 +1,73 @@
...
+import org.talend.daikon.properties.property.Property;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+import static org.talend.daikon.properties.presentation.Widget.widget;
+import static org.talend.daikon.properties.property.PropertyFactory.*;
+
+public class TestComponentProperties extends ComponentPropertiesImpl {
```
sgandon: there is a module in this repo with a full example **components-api-full-example**, please could you use it and even improve it if some things are missing.

bchen-talend: I tried components-api-full-example, but failed by maven cycle dependency problem. Do you have some suggestion? Thanks for the review!

sgandon: I would suggest to put your code into the module *https://github.com/Talend/components/tree/master/components-api-runtime-service/components-api-runtime-service*. This module may be renamed but this where we will have the rest service entry point.

francisu:
Some general comments:
1) This belongs in Daikon, not components, as it's entirely related Daikon properties and has no dependency on TCOMP. This means that a part of the components-full-example needs to be moved to Daikon as well in conjunction with this.
2) Does JSON Schema require Jackson, or can it work with json-IO.
3) Look at how this works with the Daikon serialization mechanism.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/90/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/90/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/93/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/93/)

bchen-talend:
@sgandon I did some refact and use fullexample and add some comment here, please help review, Thanks!
@francisu, for 1, yes, I hope to make it in daikon only, but there is some problem for ComponentReferenceProperties which I need to handle it for now.
for 2, I just use jackson to build a json, but unfortunately json-io do not provide these API.
for 3, ok, I will look at it after fix 1.

Here is a document and some remark

# TCOMP and JSONSchema

In TCOMP, ComponentProperties define the data struct and form/widget.

```
data struct can be defined by Property\<T\> and ComponentProperties and PresentationItem(it's for widget not data)

form type: Form.MAIN/Form.ADVANCED/Form.REFERENCE/Form.WIZARD/Custom type

widget in form could less than properties defined, treat these property/properties as hidden widget.
```

* Questions/Remarks
    * One property to one widget for jsonschema(not two more, see FullExampleProperties, it's possible to define two widget in same form, but what's mean? need to considered?) 
    * How to choice Form? It's normally to specify more than one form in one ComponentProperties and also a custom type name of Form, for now Studio has hard code, Main/Advanced for component setting, Wizard for wizard. But we'd better handle this logic in server side not front end.
    * It's possible to add Properties in widget, but do not specify the Form type, then use Main as the default?
    * It's possible to define one PresentationItem which open a new form, how to handle? 
    * Invoke trigger should with full path, means the client need to use [parent ComponetProperties name]/[current ComponentProperties name]/[Property name]/[after/before/validate]
    * Treat avro schema/data as a json string nested in jsonschema/data is it ok?

Document of [JSONSchema](http://json-schema.org/latest/json-schema-core.html) and [UISchema](https://github.com/mozilla-services/react-jsonschema-form). [Quick test page](https://mozilla-services.github.io/react-jsonschema-form/)

* TCOMP Property Type, Property\<T\>, T can be anything, most common are:
    * String
    * Boolean
    * Number(Integer/Double/Float)
    * Date
    * Enum class
    * org.apache.avro.Schema
    * List\<T\>
    * How about others?

* JSONSchema Type
    * array
    * boolean
    * integer
    * number
    * null
    * object
    * string

* Mapping between Property Type and JSONSchema Type

Property Type | JSONSchema Type
------------- | ---------------
String | string
Boolean | boolean
Long/Integer | integer
Double/Float | number
Date | string
Enum class | string
org.apache.avro.Schema | string
List\<T\> | array
ComponentProperties | object
? | null
others | ?


* TCOMP Widget Type

Type name | Property type | Comment
--------- | ------------- | -------
widget.type.default | T | default means use the proper widget based on Property type
widget.type.schema.editor | org.apache.avro.Schema | schema editor
widget.type.schema.reference | org.apache.avro.Schema | schema editor with repository selector
widget.type.name.selection.area | String only? | show a list, and a textfield, select one from the liswhich will fill the textfield or fill it by user |
widget.type.name.selection.widget | String only? | show a testfield and a button, when click the buttonshow a list like selection.area |
widget.type.component.reference | String | list components in the same job, component type defined iComponentReferenceProperties.componentType, and the user can select one of them |
widget.type.button | x | button without data
widget.type.table | ComponentProperties | table which defined by ComponentProperties, type of column'value of the table is List\<T\>, means it store the data column by column not row by row |
widget.type.hidden.text | String | for password
widget.type.file | String | file selector, how it work on Web UI? local/cluster/filesystem(hdfs/s3/..)
widget.type.enumeration | Enum class/String | list all the candidate by Enum class or possible value oString Property |
widget.type.textArea | String | multiple line of text
widget.type.jsonTable | String | use json to represent data table


* UISchema Type
    * default
    * boolean
        * checkbox[default]
        * radio
        * select
    * string
        * text[default]
        * textarea
        * password
        * color
        * email
        * url
        * file
        * date/alt-date [date not works on firefox and IE, use alt-date]
        * datetime-local/alt-datetime [same as date/alt-date]
        * select
    * number/integer
        * text[default]
        * updown
        * range
        * select
    * array
        * checkboxes
    * custom type

* UISchema functions:
    * disable
    * readonly
    * hidden
    * order
    * i18n
        * title
        * enum
        ```
        enum: [1, 2, 3],
        enumNames: ["one", "two", "three"]
        ```
    * [help](https://github.com/mozilla-services/react-jsonschema-form#help-texts)
    * [placeholders](https://github.com/mozilla-services/react-jsonschema-form#placeholders)
    
* Mapping between Widget Type and UISchema Type



Widget Type | UISchema Type
----------- | -------------
widget.type.default | default
widget.type.schema.editor | custom
widget.type.schema.reference | custom
widget.type.name.selection.area | custom
widget.type.name.selection.widget | custom
widget.type.component.reference | custom
widget.type.table | custom
widget.type.hidden.text | password
widget.type.file | file?
widget.type.enumeration | enum
widget.type.textArea | textarea
widget.type.jsonTable | custom

francisu:
@bchen-talend Thanks for your replies. Here are my answers to your questions:

Questions/Remarks
One property to one widget for jsonschema(not two more, see FullExampleProperties, it's possible to define two widget in same form, but what's mean? need to considered?)
>> Yes, it's possible that a property might be defined multiple times in a single form with different widgets. I'm not sure what the use case is exactly however, but it's allowed. Why is this a problem? You should key off of the widgets and not the properties.

How to choice Form? It's normally to specify more than one form in one ComponentProperties and also a custom type name of Form, for now Studio has hard code, Main/Advanced for component setting, Wizard for wizard. But we'd better handle this logic in server side not front end.
>> Not sure about this. And on the web side we might have even some different forms (we might have a "citizen user" form vs. a "developer" form. Probably something in the client will specify the desired form and we should have standard form times. This requires some more thought.

It's possible to add Properties in widget, but do not specify the Form type, then use Main as the default?
>> Can you given an example of how this is done. All widgets are associated with some form as far as I know. In fact, the Widget is the thing that the forms are built from (some widgets don't even have properties, like the ones that hold a PresentationItem).

It's possible to define one PresentationItem which open a new form, how to handle?
>> Yes, a button that opens a dialog for example. You see this in the SalesforceConnectionPropreties Wizard form where it opens the Advanced form. This should do the same thing in the web UI.

Invoke trigger should with full path, means the client need to use [parent ComponetProperties name]/[current ComponentProperties name]/[Property name]/[after/before/validate]
>> The PropertiesService is intended to take the Properties object by value. And then it would modify the fields in the Properties object. It needs to potentially access any of the objects in the Properties object graph (at least of any nested objects). So the Properties objects need to somehow be reconstructed before this service is called with the values from the browser [JSON Schema] (and then the JSON Schema needs to be repopulated after the call, since the service could change anything in the Properties object). 

Treat avro schema/data as a json string nested in jsonschema/data is it ok?
>> Not sure, depends on how this is used.

components-api-full-example/src/main/java/org/talend/components/fullexample/FullExampleProperties.java
```diff
@@ -42,10 +45,49 @@
...
+        public void setupLayout() {
+            Form mainForm = new Form(this, Form.MAIN);
+            mainForm.addRow(colString);
+            mainForm.addRow(colEnum);
+            mainForm.addRow(colBoolean);
+        }
+    }
+
+    public static class TableProperties extends PropertiesImpl {
+        private static final TypeLiteral<List<String>> LIST_STRING_TYPE = new TypeLiteral<List<String>>() {// empty
```
talend-glorieux: Why the `// empty` comment? Is it used in any way by the JavaDoc?

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONResolver.java
```diff
@@ -0,0 +1,95 @@
...
+            IllegalAccessException, InvocationTargetException, InstantiationException {
+        JsonNode jsonSchema = mapper.readTree(jsonSchemaStr);
+        JsonNode jsonData = mapper.readTree(jsonDataStr);
+
+        if (jsonSchema.isObject()) {
+            JsonNode classNameNode = jsonSchema.get(JSONSchemaMeta.TAG_ID);
+            jsonSchema = jsonSchema.get(JSONSchemaMeta.TAG_PROPERTIES);
+            Class<?> aClass = Class.forName(classNameNode.textValue());
+            if (Properties.class.isAssignableFrom(aClass)) {
+                if(cProperties == null) {
```
talend-glorieux: Missing space between if and opening parenthesis.

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/UISchemaGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+     * @param cProperties
+     * @return UISchema
+     */
+    private ObjectNode processTPropertiesWidget(Properties cProperties) {
+        Form mainForm = cProperties.getForm(Form.MAIN);
+        Form advancedForm = cProperties.getForm(Form.ADVANCED);
+        return processTPropertiesWidget(mainForm, advancedForm);
+    }
+
+    private ObjectNode processTPropertiesWidget(Form... forms) {
```
talend-glorieux: The cyclomatic complexity of that method seems pretty high.

components-api-runtime-service/components-api-runtime-service/src/test/resources/org/talend/components/runtimeservice/util/json/FullExampleProperties.json
```diff
@@ -0,0 +1 @@
...ullexample.FullExampleProperties$TableProperties","properties":{"colListString":{"type":"arrays","items":{"type":"string"}},"colListEnum":{"type":"arrays","items":{"type":"string","enum":["FOO","BAR"]}},"colListBoolean":{"type":"arrays","items":{"type":"boolean"}}}},"commonProp":{"type":"object","id":"org.talend.components.fullexample.FullExampleProperties$CommonProperties","properties":{"colString":{"type":"string"},"colEnum":{"type":"string","enum":["FOO","BAR"]},"colBoolean":{"type":"boolean"}}}},"required":["integerProp","dateProp"]},"jsondata":{"stringProp":"initialValue","hideStringPropProp":false,"tableProp":{},"commonProp":{"colBoolean":false}},"uischema":{"schema":{"ui:field":"schema"},"commonProp":{"ui:order":["colString","colEnum","colBoolean"]},"hiddenTextProp":{"ui:widget":"password"},"filepathProp":{"ui:widget":"file"},"tableProp":{"ui:order":["colListString","colListEnum","colListBoolean"],"ui:field":"table"},"textareaProp":{"ui:widget":"textarea"},"ui:order":["stringProp","schema","multipleSelectionProp","commonProp","hiddenTextProp","filepathProp","integerProp","dateProp","tableProp","textareaProp","hideStringPropProp"],"hideStringPropProp":{"ui:widget":"hidden"}}}
```
talend-glorieux: Is this file generated? If not there is a missing newline at the end of the file.

components-webtest/src/main/java/org/talend/components/webtest/JsonSchema2HttpMessageConverter.java
```diff
@@ -0,0 +1,84 @@
...
+     * @param t Object to serialize
+     * @param outputMessage the HTTP output message to write to
+     */
+    @Override
+    protected void writeInternal(Object t, HttpOutputMessage outputMessage) throws IOException, HttpMessageNotWritableException {
+        OutputStreamWriter outputStreamWriter = new OutputStreamWriter(outputMessage.getBody());
+        String objectToJson = null;
+        if(Properties.class.isAssignableFrom(t.getClass())){
+            objectToJson = JSONUtil.toJson((Properties)t, true);
+        }else {
```
talend-glorieux: Small formatting issue on the if and else.

components-webtest/src/main/java/org/talend/components/webtest/JsonSchema2HttpMessageConverter.java
```diff
@@ -0,0 +1,84 @@
...
+            try {
+                return JSONUtil.fromJson(inputMessage.getBody());
+            } catch (Exception e) {
+                throw new IOException(e.getMessage());
+            }
+        }else {
+            return JsonReader.jsonToJava(inputMessage.getBody(), null);
+        }
+    }
+}
```
talend-glorieux: Missing newline at the end of the file.

components-webtest/src/main/java/org/talend/components/webtest/JsonSchema2HttpMessageConverter.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    protected Object readInternal(Class<? extends Object> clazz, HttpInputMessage inputMessage)
+            throws IOException, HttpMessageNotReadableException {
+        if(Properties.class.isAssignableFrom(clazz)){
+            try {
+                return JSONUtil.fromJson(inputMessage.getBody());
+            } catch (Exception e) {
+                throw new IOException(e.getMessage());
+            }
+        }else {
```
talend-glorieux: Formatting issues on the if and else.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/106/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/106/)

components-api-full-example/src/main/java/org/talend/components/fullexample/FullExampleProperties.java
sgandon: We have an eclipse check for empty block that shows an warning if nothing is in the block, expecting it to be commented why we have an empty block.
I agree that the comment should be more explicit.

components-api-full-example/src/main/java/org/talend/components/fullexample/FullExampleProperties.java
```diff
@@ -29,7 +32,7 @@
      */
     static final String POPUP_FORM_NAME = "popup";
 
-    public static class TableProperties extends PropertiesImpl {
+    public static class CommonProperties extends PropertiesImpl {
```
sgandon: Why did you rename it? it really looks like a Table properties to me ?

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/UISchemaGenerator.java
sgandon: I agree, please could you split it into multiple methods. thanks.

components-api-full-example/src/main/java/org/talend/components/fullexample/FullExampleProperties.java
talend-glorieux: OK. Sounds like a hack to me. 
Is there a better solution like fixing the linter or removing the need for empty blocks?

sgandon: There is not way to avoid that empty block.
And removing this check from the linter would be very bad because you would accept any empty block without any justification.
This is a way to justify what you are doing.
It is exactly the same reason why I don't like code like this : 
```
if (something){
    do something
}

```
and prefer such code
```
if (something){
    do something
}//else not something so ...

```
That way you know that the else case has been thought about and this is not an *else* case that has been forgotten.

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONResolver.java
```diff
@@ -0,0 +1,95 @@
...
+            if (Float.class.getName().equals(javaType)) {
+                return dataNode.numberValue().floatValue();
+            }
+            return dataNode.numberValue();
+        } else if (JSONSchemaMeta.TYPE_BOOLEAN.equals(TYPE_MAPPING.get(javaType))) {
+            return dataNode.booleanValue();
+        } else if (Schema.class.getName().equals(javaType)) {
+            return new Schema.Parser().parse(dataNode.textValue());
+        } else if (findClass(javaType).isEnum()) {
+            return Enum.valueOf(findClass(javaType), dataNode.textValue());
```
RyanSkraba: Unfortunate to do two `findClass` calls so close together... but see my comment on `findClass` in the JsonBaseTool.

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONBaseTool.java
```diff
@@ -0,0 +1,123 @@
...
+    protected ObjectNode createSchema() {
+        return mapper.createObjectNode();
+    }
+
+    /**
+     * the className is generated by TypeUtils.toString(), this method is try to find the class by this string but may failed
+     * @param className
+     * @return
+     */
+    protected Class findClass(String className) {
```
RyanSkraba: This is only ever used on a className that is coming from `property.getType()`, sometimes extracting the inner class if the type is "List<xxx>", and it's only used for enum purposes.  How about doing a `instanceof EnumProperty` instead, since it's only used for enum purposes?

We don't want to be sending and instantiating classes from class names in data received over a REST API!

There is no way to get the type from a Property without going through this process, even if we have the compiled XxxxProperties in our hands?  

@sgandon : I do not know why we aren't keeping the `TypeLiteral`, so we can just check it directly!

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONBaseTool.java
```diff
@@ -0,0 +1,123 @@
...
+                throw new RuntimeException(e);
+            }
+            className = className.substring(0, lastPosition) + "$" + className
+                    .substring(lastPosition + 1);
+            return findClass(className);
+        }
+        return aClass;
+    }
+
+    protected List<Property> listTProperty(Properties cProperties) {
```
RyanSkraba: This should be moved directly to the Properties class!

RyanSkraba: Probably already renamed `getSubProperty` or `getPropertyMembers`.

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONBaseTool.java
```diff
@@ -0,0 +1,123 @@
...
+                    propertyList.add((Property) field.get(cProperties));
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+        }
+        return propertyList;
+    }
+
+    protected List<Properties> listTProperties(Properties cProperties) {
```
RyanSkraba: Also directly in Properties, probably should be renamed `getSubProperties` or `getPropertiesMembers`.

In general, where you say `TProperty` or `TProperties` in this pull request, it would be better to say SubProperty/SubProperties or SubProperty/NestedProperties.

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONBaseTool.java
```diff
@@ -0,0 +1,123 @@
...
+        WIDGET_MAPPING.put(Widget.TEXT_AREA_WIDGET_TYPE, UISchemaMeta.TYPE_TEXT_AREA);
+        // null means use the default
+        // WIDGET_MAPPING.put(Widget.DEFAULT_WIDGET_TYPE, null);
+        // WIDGET_MAPPING.put(Widget.NAME_SELECTION_AREA_WIDGET_TYPE, null);
+        // WIDGET_MAPPING.put(Widget.NAME_SELECTION_REFERENCE_WIDGET_TYPE, null);
+        // WIDGET_MAPPING.put(Widget.COMPONENT_REFERENCE_WIDGET_TYPE, null);
+        // WIDGET_MAPPING.put(Widget.ENUMERATION_WIDGET_TYPE, null);
+    }
+
+    protected ObjectNode createSchema() {
```
RyanSkraba: All instances can be replaced by `JsonNodeFactory.instance.createObjectNode()`, since it's never actually used to do object mapping.

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONDataGenerator.java
```diff
@@ -0,0 +1,58 @@
+package org.talend.components.runtimeservice.util.json;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.apache.avro.Schema;
+import org.talend.daikon.properties.Properties;
+import org.talend.daikon.properties.property.Property;
+
+import java.util.List;
+
+public class JSONDataGenerator extends JSONBaseTool {
```
RyanSkraba: Bof: `JsonDataGenerator` and `generateJsonData(...)`.

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONResolver.java
```diff
@@ -0,0 +1,95 @@
...
+import org.talend.daikon.properties.property.Property;
+
+import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Modifier;
+import java.util.List;
+
+public class JSONResolver extends JSONBaseTool {
```
RyanSkraba: And `JsonResolver`  I'd be tempted to put this in the same class as `JsonDataGenerator` since it is the same process in the other direction!

components-api-full-example/src/main/java/org/talend/components/fullexample/FullExampleProperties.java
```diff
@@ -65,12 +107,24 @@ public TableProperties(String name) {
     /** checking {@link WidgetType#TABLE} */
     public final TableProperties tableProp = new TableProperties("tableProp");
```
sgandon: Yes it is the right way, each columns list will contain the row data.

components-api-full-example/src/main/java/org/talend/components/fullexample/FullExampleProperties.java
sgandon: right, I did not see the table below.

components-api-full-example/src/main/java/org/talend/components/fullexample/FullExampleProperties.java
```diff
@@ -113,9 +167,14 @@ public void setupLayout() {
         mainForm.addRow(widget(showNewForm).setWidgetType(Widget.BUTTON_WIDGET_TYPE));
         Form popUpForm = new Form(this, POPUP_FORM_NAME);
```
sgandon: what is the problem here, this is how forms are linked to the Properties instances ?

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONBaseTool.java
```diff
@@ -0,0 +1,123 @@
...
+import java.lang.reflect.Field;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.*;
+
+/**
+ * Generate json-schema and ui-schema from {@ComponentProperties},
+ * and only consider {@Property} and {@Properties} inside.
+ */
+public class JSONBaseTool {
```
sgandon: I agree

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONBaseTool.java
sgandon: If I recall well, the TypeLiteral was not serializing properly (with json-io), this is why I did not keep it. We may rethink about it.

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONBaseTool.java
sgandon: this will be difficult to change the Properties interface, I really woulld suggest you just remove those 2 methods and whenever you need to access all the property or properties, please use the *org.talend.daikon.properties.property.PropertyVisitor* and the *org.talend.daikon.properties.PropertiesVisitor*. See an example here : org.talend.components.api.properties.ComponentPropertiesImpl.updateNestedProperties(Properties)

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONBaseTool.java
sgandon: See my comment above

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONSchemaMeta.java
```diff
@@ -0,0 +1,22 @@
+package org.talend.components.runtimeservice.util.json;
+
+/**
+ * Metadata of JSONSchema, refer to http://json-schema.org/
+ */
+public class JSONSchemaMeta {
+    public static final String TAG_TYPE = "type";
+    public static final String TAG_ID = "id";
```
sgandon: yes indeed you'll need some mechanism for deserializing json to java to make sure you end-up with the right instance.

components-api-full-example/src/main/java/org/talend/components/fullexample/FullExampleProperties.java
RyanSkraba: This is a working implementation, but it is unexpected for me that the constructor of the Form changes the Properties that was passed in!  This is unrelated to the PR in any case.  I didn't make a JIRA :)

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONSchemaMeta.java
bchen-talend: Definition name is not a part of ComponentProperties.
Do you think we need a header/meta which include `family name`,  `definition name` and `version`.
family name and the definition name are the unique id, and version is for migration task

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONBaseTool.java
bchen-talend: for `Property<List<ColEnum>>`, can not use `instanceof EnumProperty` instead, unless we add `ListEnumProperty`?

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONBaseTool.java
bchen-talend: PropertyVisitor and PropertiesVisitor are great way! I may can use them to refact JsonDataGenerator and JsonSchemaGenerator. 
But I need to keep getSubProperty/getSubProperties(I have change the name as Ryan suggest) for UiSchemaGenerator

sgandon: why ?

bchen-talend: For UiSchemaGenerator, I need a visitor for start from From no ComponentProperties, for the current UiSchemaGenerator will handle nested CompnoentProperties one level to another deep level and I need to caculate which Property in this level do not in Form and mark hidden for them. So I need a way to get current sub properties/property.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/189/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/189/)

components-api-runtime-service/components-api-runtime-service/src/main/java/org/talend/components/runtimeservice/util/json/JSONSchemaMeta.java
bchen-talend: I will keep the classname here for Serializer/Deserializer.
And will do some extra job on REST server side to remove the classname information and add definition name. (because there is no way to get Definition in ComponentProperties)
For the version, it's no need after looking at SerializerDeserializer in Daikon, it provide a way to set version and how to do migration in ComponentProperties, need to make it work in this lib together.

bchen-talend:
failed by maven cycle dependency problem again, after merge master, where I hold this code then?

sgandon:
After talking with francis, we think this should code should belong to daikon, do you think it is possible regarding the dependencies that you have ?

bchen-talend:
closed as it move to daikon

PR #249, opened 2016-09-22
Feature/6.3/tdi 36957 file delimited

jzhao-talend:


jzhao-talend:
This is the first pull request for delimited components.
Need continue with TTP system tuj test result.
And add more test to cover more code.

jzhao-talend:
Because of we need to integrate this ASAP. SO just ignore the failed build which caused by timezone
Would fix them tomorrow.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/6/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/6/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/7/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/7/)

sgandon:
Sorry, but build must pass before merging it to master.
We will wait until you fix it tomorrow to merge it.

gonchar-ivan:
General comment:
low comments and javadocs

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/11/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/11/)

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedWriter.java
```diff
@@ -0,0 +1,230 @@
...
+    FileDelimitedWriter(FileDelimitedWriteOperation writeOperation, RuntimeContainer container) {
+        this.writeOperation = writeOperation;
+        this.container = container;
+        sink = writeOperation.getSink();
+        props = sink.getOutputProperties();
+        outputRuntime = new FileOutputDelimitedRuntime(props);
+    }
+
+    @Override
+    public void open(String uId) throws IOException {
```
gonchar-ivan: rather complex logic...

in this open() you do 3 steps:
1. Retrieve recordSchema   (I think better to retrieve recordSchema in construcotr.  open(), write(), close() workflow could be executed several times, but properties won't be changed between each execution. So, here in open should be initialized those things which should be reset after each execution ).

2. Create/get Writer. Two Writers are possible: CVS and IO. 
There a lot of if (csvWriter != null) further in the code.
It is better to handle such things using inheritance - 2 separate FileDelimitedWriter classes
FileDelimitedCvsWriter and FileDelimitedIOWriter

3. write header... also depends on type of writer

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/DelimitedReader.java
```diff
@@ -0,0 +1,107 @@
...
+            if (fid != null) {
+                fid.close();
+            }
+        }
+        LOGGER.debug("close: " + properties.fileName.getStringValue());
+    }
+
+    protected void retrieveValues() throws IOException {
+        if (inputRuntime.schemaIsDynamic) {
+            values = new String[fid.getColumnsCountOfCurrentRow()];
```
chmyga: Why do we return a new String array every time? I thought we have to init schema only once, if it is dynamic. How can we have different column number in different lines? Is it a correct behavior?

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileCSVReader.java
```diff
@@ -0,0 +1,133 @@
...
+import org.talend.components.filedelimited.tfileinputdelimited.TFileInputDelimitedProperties;
+
+import com.talend.csv.CSVReader;
+
+/*
+ * CSV mode reader
+ */
+public class FileCSVReader extends FileDelimitedReader {
+
+    private static final long serialVersionUID = 1L;
```
chmyga: use generated serial version.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileCSVReader.java
```diff
@@ -0,0 +1,133 @@
...
+        boolean startAble = false;
+        try {
+            inputRuntime.init();
+            LOGGER.debug("open: " + properties.fileName.getStringValue());
+            csvReader = inputRuntime.getCsvReader();
+            currentLine = inputRuntime.currentLine;
+            startAble = inputRuntime.limit != 0 && csvReader != null && csvReader.readNext();
+            if (startAble) {
+                values = csvReader.getValues();
+                retrieveValues();
```
chmyga: it looks like here we do the same action twice - we get values from csvReader - once directly, and once in the retrieveValues() method. Do we need to do that?

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileCSVReader.java
```diff
@@ -0,0 +1,133 @@
...
+                startAble = checkLimit();
+                currentIndexRecord = ((DelimitedAdaptorFactory) getFactory()).convertToAvro(values);
+            }
+        } catch (IOException e) {
+            if (properties.dieOnError.getValue()) {
+                throw e;
+            } else {
+                // TODO Meed junit test
+                startAble = advance();
+            }
```
chmyga: If the component shouldn't die on error, and we have two errors in a row, it will die anyway. It should be done in some other way, I think.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileCSVReader.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public boolean advance() throws IOException {
+        boolean isContinue = false;
+        try {
+            isContinue = csvReader.readNext();
+            if (!isContinue) {
+                if (properties.uncompress.getValue()) {
+                    csvReader = inputRuntime.getCsvReader();
+                    isContinue = inputRuntime.limit != 0 && csvReader != null && csvReader.readNext();
+                    currentLine = inputRuntime.currentLine;
```
chmyga: better to use getters for those fields(inputRuntime.limit, inputRuntime.currentLine, etc.).

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileInputDelimitedRuntime.java
```diff
@@ -0,0 +1,373 @@
...
+import org.talend.components.common.EncodingTypeProperties;
+import org.talend.components.common.runtime.FileRuntimeHelper;
+import org.talend.components.filedelimited.tfileinputdelimited.TFileInputDelimitedProperties;
+import org.talend.daikon.avro.AvroUtils;
+import org.talend.fileprocess.FileInputDelimited;
+
+import com.google.gson.Gson;
+import com.talend.csv.CSVReader;
+
+public class FileInputDelimitedRuntime {
```
chmyga: Not sure that FileInputDelimitedRuntime is a good name for that class. Runtime is something like Source/Sink/ComponentRuntime. It's something that is called in runtime. Why do you call it like that? Is this class required at all? Common methods for readers can be moved to the Reader class, and reader-specific(like getCsvReader(), getFileInputReader()) can be moved to the specific readers. I don't really understand the real purpose of this class. I can't see it is being used in any different class except FileDelimitedReader.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedReader.java
```diff
@@ -0,0 +1,91 @@
...
+    protected final List<IndexedRecord> rejectedWrites = new ArrayList<>();
+
+    protected int dataCount;
+
+    public FileDelimitedReader(RuntimeContainer container, BoundedSource source, TFileInputDelimitedProperties properties) {
+        super(source);
+        this.container = container;
+        this.properties = properties;
+        schema = properties.main.schema.getValue();
+        inputRuntime = new FileInputDelimitedRuntime(properties);
```
chmyga: Not sure it is a good decision to use this class. But it is commented under FileInputDelimitedRuntime class.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedSource.java
```diff
@@ -0,0 +1,97 @@
...
+    public ValidationResult validate(RuntimeContainer container) {
+        ValidationResult vr = super.validate(container);
+        // also check that the properties is the right type
+        if (vr.getStatus() != ValidationResult.Result.ERROR) {
+            if (!(properties instanceof TFileInputDelimitedProperties)) {
+                return new ValidationResult().setStatus(ValidationResult.Result.ERROR)
+                        .setMessage("properties should be of type :" + TFileInputDelimitedProperties.class.getCanonicalName());
+            }
+        }
+        Object fileOrStream = ((FileDelimitedProperties) properties).fileName.getValue();
```
chmyga: if we have an error in super.validate() method, we shouldn't go further. We should return error result from the superclass, as those should be basic results.

components-common/src/main/java/org/talend/components/common/BasedOnSchemaTable.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.apache.commons.lang3.reflect.TypeLiteral;
+import org.talend.components.api.properties.ComponentPropertiesImpl;
+import org.talend.daikon.properties.presentation.Form;
+import org.talend.daikon.properties.property.Property;
+
+import static org.talend.daikon.properties.property.PropertyFactory.newProperty;
+
+public class BasedOnSchemaTable extends ComponentPropertiesImpl {
+
```
jzhao-talend: Yes, We gave a lot of table which generated based schema
https://github.com/Talend/tdi-studio-se/blob/master/main/plugins/org.talend.designer.components.localprovider/components/tIterateToFlow/tIterateToFlow_java.xml#L54
This kind of table need this.

components-common/src/main/java/org/talend/components/common/EncodingTypeProperties.java
```diff
@@ -0,0 +1,64 @@
...
+    public EncodingTypeProperties(String name) {
+        super(name);
+    }
+
+    @Override
+    public void setupProperties() {
+        super.setupProperties();
+        List<String> defaultEncodings = getDefaultEncodings();
+        if (defaultEncodings.size() > 0) {
+            encodingType.setPossibleValues(getProperties());
```
jzhao-talend: Yes. Should be "defaultEncodings".
It's strange. It can be show correct in the UI. I would check that.
Thanks.

components-common/src/main/java/org/talend/components/common/ValuesTrimPropertis.java
```diff
@@ -0,0 +1,64 @@
...
+import org.talend.components.api.properties.ComponentPropertiesImpl;
+import org.talend.daikon.properties.presentation.Form;
+import org.talend.daikon.properties.presentation.Widget;
+import org.talend.daikon.properties.property.Property;
+
+import static org.talend.daikon.properties.presentation.Widget.widget;
+import static org.talend.daikon.properties.property.PropertyFactory.newBoolean;
+
+public class ValuesTrimPropertis extends ComponentPropertiesImpl {
+
```
jzhao-talend: Got it. thanks.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileCSVReader.java
jzhao-talend: This part need to recheck with reject.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileCSVReader.java
jzhao-talend: Need to refactor those two runtime class

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedReader.java
jzhao-talend: Need to refactor this.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedSource.java
jzhao-talend: yes. you are right.
Also need junit for this.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedWriter.java
```diff
@@ -0,0 +1,230 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(FileDelimitedWriter.class);
+
+    private final FileDelimitedSink sink;
+
+    private IndexedRecordConverter<IndexedRecord, IndexedRecord> factory;
+
+    private WriteOperation<Result> writeOperation;
+
+    private final RuntimeContainer container;
```
jzhao-talend: Thanks. would remove it.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedWriter.java
```diff
@@ -0,0 +1,230 @@
...
+    private CSVWriter csvWriter;
+
+    private java.io.Writer writer;
+
+    private int currentRowNo = 0;
+
+    FileDelimitedWriter(FileDelimitedWriteOperation writeOperation, RuntimeContainer container) {
+        this.writeOperation = writeOperation;
+        this.container = container;
+        sink = writeOperation.getSink();
```
jzhao-talend: you are right.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedWriter.java
chmyga: I totally agree with Ivan.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileInputDelimitedRuntime.java
```diff
@@ -0,0 +1,373 @@
...
+            }
+        } else {
+            if (sourceIsStream) {
+                checkFooter();
+                csvReader = new CSVReader((java.io.InputStream) fileNameOrStream, fieldSeparator, encoding);
+            } else {
+                csvReader = new CSVReader(new java.io.BufferedReader(
+                        new java.io.InputStreamReader(new java.io.FileInputStream(String.valueOf(fileNameOrStream)), encoding)),
+                        fieldSeparator);
+            }
```
chmyga: All readers and streams should be closed once they aren't needed anymore. If you don't close them, it might result in memory leaks.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileInputDelimitedRuntime.java
```diff
@@ -0,0 +1,373 @@
...
+
+    public void checkFooter() throws IOException {
+        if (footer > 0) {
+            throw new IOException("When the input source is a stream,footer shouldn't be bigger than 0.");
+        }
+    }
+
+    // Preview data and guess the columns
+    public String previewData(int maxRowsToPreview) throws IOException {
+        init();
```
chmyga: As it was already told before - the logic is too heavy and unreadable. Better extract common methods to common FileInputReader and specific methods, like this one, to the specific inherited classes(like FileCsvReader).

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedWriter.java
jzhao-talend: 1. About retrieve records schema:
If design schema not include dynamic field.We can get record schema (same with deign schema).
If include dynamic field, we need get the record schema from IndexRecord.

2. About the writers, now I just get the logic from javajet.
I would improve them to make them more clear

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedWriter.java
```diff
@@ -0,0 +1,230 @@
...
+            } else {
+                if (writer != null) {
+                    writer.flush();
+                    outputRuntime.streamWriter.flush();
+                    writer.close();
+                }
+            }
+        }
+
+        result.successCount = result.totalCount;
```
jzhao-talend: yes. you are right. This is written by another colleague.
I didn't have a detail check. 
Would remove it.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileInputDelimitedRuntime.java
jzhao-talend: I also consider to refactor this. Would do it to make it more clear.

components-filedelimited/src/test/java/org/talend/components/filedelimited/runtime/FileDelimitedReaderTestIT.java
```diff
@@ -0,0 +1,339 @@
...
+
+        // when specified header line include empty field value like ";;;;"
+        // Avro can be set field name empty. But we can do that before
+        properties.removeEmptyRow.setValue(false);
+        // properties.header.setValue(1);
+        // testInputDynamic(properties, 9);
+    }
+
+    protected void testBasicInput(TFileInputDelimitedProperties properties, int count, boolean previewData) throws Throwable {
+        if (previewData) {
```
chmyga: why don't you use two methods - testBasicInputPreview(TFileInputDelimitedProperties properties, int count) and testBasicInput(TFileInputDelimitedProperties properties, int count)? this doesn't make much sense to split the code with a boolean variable like that. Those two parts of code don't have any common parts, so they can be separated easily and the code would be more readable - less lines of code in the method and people, who read the method don't have to keep all the booleans in their heads.

components-filedelimited/src/test/java/org/talend/components/filedelimited/runtime/FileDelimitedReaderTestIT.java
```diff
@@ -0,0 +1,339 @@
...
+        assertTrue(records.get(0).get(2) instanceof byte[]);
+        assertTrue(records.get(0).get(3) instanceof Character);
+        assertTrue(records.get(0).get(4) instanceof Long);
+        assertTrue(records.get(0).get(5) instanceof Double);
+        assertTrue(records.get(0).get(6) instanceof Float);
+        assertTrue(records.get(0).get(7) instanceof BigDecimal);
+        assertTrue(records.get(0).get(8) instanceof Integer);
+        assertTrue(records.get(0).get(9) instanceof Long);
+        assertTrue(records.get(0).get(10) instanceof byte[]);
+        if (properties.header.getValue() == 1) {
```
chmyga: Try not to use such constant values in your tests or document them. If someone will add some test with "header.getValue()" not equal to 1 or 3, the test might not work as it is expected. Such things must be documented.

components-filedelimited/src/test/java/org/talend/components/filedelimited/runtime/FileDelimitedWriterTestIT.java
```diff
@@ -0,0 +1,234 @@
...
+            assertTrue(FileRuntimeHelper.compareInTextMode(outputFile, refFilePath, getEncoding(properties.encoding)));
+            assertTrue(deleteFile(properties.fileName.getStringValue()));
+        }
+
+    }
+
+    protected void testOutputDynamic(boolean isCsvMode) throws Throwable {
+        String resources = getResourceFolder();
+        String outputFile = resources + "/out/output_delimited_dynamic.csv";
+        // TODO need to implement
```
chmyga: need to implement. did you forget about it, or do you want to implement it later?

pom.xml
```diff
@@ -20,11 +20,10 @@
 		<module>components-webtest</module>
 		<module>components-osgitest</module>
 		<module>components-api-proptester</module>
-		<module>components-splunk</module>
-		<module>components-jira</module>
```
chmyga: why did you remove those two modules? was there a requirement to remove them?

components-filedelimited/src/main/java/org/talend/components/filedelimited/wizard/FileDelimitedWizardProperties.java
```diff
@@ -0,0 +1,135 @@
...
+    public ValidationResult validatePreview() {
+        ValidationResult vr = new ValidationResult();
+        try {
+            if (!StringUtils.isEmpty(fileName.getStringValue())) {
+                File file = new File(fileName.getStringValue());
+                if (!file.exists() || !file.isFile()) {
+                    throw new IOException("File \"" + fileName.getStringValue() + "is not found.");
+                }
+                Map<String, Schema> dataAndSchema = FileDelimitedSource.previewData(null, this, 200);
+                for (String jsonData : dataAndSchema.keySet()) {
```
chmyga: you can use an entrySet() instead of keySet(). That way you won't need to get a Schema by the key, and you can get the key and the value from the entry.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileSourceOrSink.java
```diff
@@ -0,0 +1,102 @@
...
+            for (int columnIndex = 0; columnIndex < fieldsSize; columnIndex++) {
+                int columnLength = columnsLength.get(columnIndex);
+                String columnName = null;
+                if (columnsName != null && columnIndex < columnsName.size()) {
+                    columnName = columnsName.get(columnIndex);
+                } else {
+                    columnName = "Column" + columnIndex;
+                }
+                // TODO guess data type
+                Schema.Field avroField = new Schema.Field(columnName, AvroUtils._string(), null, defaultValue);
```
jzhao-talend: Here need to continue to implement get the field schema base on the field value.

components-filedelimited/src/main/java/org/talend/components/filedelimited/wizard/FileDelimitedWizardProperties.java
```diff
@@ -0,0 +1,135 @@
...
+    // TODO check "Format"
+
+    @Override
+    public void setupProperties() {
+        super.setupProperties();
+        encoding.encodingType.setPossibleValues(Arrays.asList(ENCODING_TYPE_UTF_8, "UTF-16", "UTF-16LE", "UTF-16BE", "UTF-7",
+                "ISO-8859-1", "ISO-8859-2", "ISO-8859-3", "ISO-8859-4", "ISO-8859-5", "ISO-8859-6", "ISO-8859-7", "ISO-8859-8",
+                "ISO-8859-9", "ISO-8859-10", ENCODING_TYPE_ISO_8859_15, "windows-1252", "BIG5", "GB18030", "GB2312", "EUC_CN",
+                ENCODING_TYPE_CUSTOM));
+        encoding.encodingType.setValue(ENCODING_TYPE_UTF_8);
```
jzhao-talend: agree with you.

pom.xml
jzhao-talend: I don't know when it was removed. I would add them back ASAP.

components-filedelimited/src/test/java/org/talend/components/filedelimited/runtime/FileDelimitedWriterTestIT.java
jzhao-talend: Implement it later.

components-filedelimited/src/test/java/org/talend/components/filedelimited/runtime/FileDelimitedReaderTestIT.java
jzhao-talend: you are right.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/36/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/36/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/38/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/38/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/55/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/55/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/57/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/57/)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/61/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/61/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/62/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/62/)

PR #252, opened 2016-09-27
Bugfix/6.3/tdi 37525 fix delimited issue

jzhao-talend:


gonchar-ivan:
It is related to
https://jira.talendforge.org/browse/TDI-37525

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedWriter.java
```diff
@@ -81,16 +81,16 @@ public void write(Object datum) throws IOException {
         if (datum == null) {
             return;
         }
-
-        IndexedRecord inputRecord = getFactory(datum).convertToAvro((IndexedRecord) datum);
```
gonchar-ivan: Why you have moved after schema getting?

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/78/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/78/)

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedWriter.java
jzhao-talend: For schema included dynamic, we need write the header firstly based on the schema.
And if "recordSchema==null" in write method means components schema include dynamic field.
Need get schema from the input record firstly

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedWriter.java
```diff
@@ -81,16 +81,16 @@ public void write(Object datum) throws IOException {
...
         if (recordSchema == null) {
-            recordSchema = inputRecord.getSchema();
+            recordSchema = ((IndexedRecord) datum).getSchema();
             if (csvWriter != null) {
                 outputRuntime.writeHeader(csvWriter, recordSchema);
             } else {
                 outputRuntime.writeHeader(writer, recordSchema);
             }
         }
+        IndexedRecord inputRecord = getFactory(datum).convertToAvro((IndexedRecord) datum);
```
gonchar-ivan: You should do it before
if (recordSchema == null) {
}
because datum could be of any type. Not only IndexedRecord.
That's why we need to call 
getFactory(datum).convertToAvro((IndexedRecord) datum);

It is required to convert Object datum to IndexedRecord.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileSourceOrSink.java
```diff
@@ -82,21 +84,70 @@ public static Schema getSchema(String schemaName, List<String> columnsName, List
...
+     * 
+     * @param columnsName All columns name which read from file
+     * @param schemaName Specify the schema name
+     * @param designSchema The design schema of current component
+     * @return The build schema include all fields
+     */
+    public static Schema getDynamicSchema(String[] columnsName, String schemaName, Schema designSchema) {
+        if (columnsName != null && columnsName.length > 0) {
+            // FIXME can we use "DiSchemaConstants" here?
+            String dynamicPosProp = designSchema.getProp(DiSchemaConstants.TALEND6_DYNAMIC_COLUMN_POSITION);
```
gonchar-ivan: Generally, it is prohibited to use DiSchemaConstants class in components.
org.talend.daikon.avro.SchemaConstants should be used instead
SchemaConstants.INCLUDE_ALL_FIELDS constant should be used to mark dynamic column.

However, I'm not  sure there is already solution for it.
Seems, you use it because of migration issue.

gonchar-ivan:
@jzhao-talend , you can merge it. However, there are 2 things we need to solve later:
1. Remove DiSchemaConstants usage
2. Move datum to IndexedRecord conversion before schema retrieving

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedWriter.java
gonchar-ivan: ok. I see getFactory() method uses recordSchema. It sets it to FileDelimitedIndexedRecordConverter.

But there is an error.
@jzhao-talend , could you, please, read this wiki
https://github.com/Talend/components/wiki/6.-Component-Runtime#emitting-data-from-input-components

Short description according wiki page:
1. You don't need implement IndexedRecordConverter at all if your Input component emits instances of IndexedRecord. 
You need to implement IndexedRecordConverter if your Input component emits data of specific type. (for example, SObject in case of Salesforce) Also you need to register it in AvroRegistry.

2. getFactory() should call AvroRegistry.getConverter(datum.getClass()). (It is not factory. It is converter. Better to name this method getConverter()) 
You should try to convert incoming datum with your own converter. You don't know all possible datum types and you don't know how to convert it to IndexedRecord

gonchar-ivan: So, here you will have ClassCastException if incoming datum won't be instance of IndexedRecord

gonchar-ivan: Also you are not use datum parameter in getFactory() method

jzhao-talend: Thanks Ivan, I would check it.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileSourceOrSink.java
jzhao-talend: I need to know the position of dynamic column.
So I use "DiSchemaConstants.TALEND6_DYNAMIC_COLUMN_POSITION" here.
**Should I created a jira for daikon to add this kind of constants to SchemaConstants ?**

Why can't I generate all field directly:
Because of we get all field values from file are string . We can't know the type
For that not included in dynamic, we need to know the the type in design schema.
For dynamic field, we would set default type "String"
So we need to know the dynamic column position to know which columns are dynamic column included and not included.

PR #253, opened 2016-09-27
Make osgi tests work for each components

sgandon:
before the module **components-osgitest** was handling all OSGI tests, and components that wanted to test their osgi bundle have to add a test in this bundle.
This PR give the possibility to the component developer to test their bundle inside the component module.
The module **components-osgitest** is still required but now it contains the basic definition for the pax-exam framework to launch the component api service. So it will be required as a dependency for all components that want to do OSGI tests.
*Salesforce*, *api-common* and the archetype have been updated to launch their own OSGI tests.


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/82/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/82/)

francisu:
I'm good with this; no problems as is. 

My only comment is the module names associated with the all of the multi-runtime test modules are not according to the standard (since you had a commit to clean up those, I'm addressing this here). All of those module names should start with components-api-test to make it clear they are both part of the components-api and they are test modules. If you want to handle this in a separate pull request, that's fine. https://jira.talendforge.org/browse/TCOMP-307

One thing that's missing though is Spring and OSGi tests for most components. What would really be nice is if those were completely automated (a separate class was not necessary at all, and perhaps generated). This is something that could be done pretty easily with Gradle I think. We can have some declarative means of saying which OSGi bundles are required as part of the Gradle definition of the component. But again, this is the next step. https://jira.talendforge.org/browse/TCOMP-306

francisu:
(Did I do the review labeling stuff right?)

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/139/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/139/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/140/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/140/)

PR #260, opened 2016-10-10
fix (TDI-37525): fix delimited issue with checking fails tujs in TTP 

jzhao-talend:

1. filedelimited write decimal precision
2. "check date" doesn't work in filedelimited read and parse date
3. number decode  doesn't work in filedelimited read and parse number ("013" / "0X13" /..)
4. short type converter is missed
5. wrong  class name "ValuesTrimPropertis" 
6. number string with thousand separator "18,888" can't be read
7. field value trim function doesn't work filedelimited input
8. filedelimited input trim/decode table can't be saved

**What is the new behavior?**

fix above issue

- [ ] Yes
- [x] No

**Other information**:
Only one commit message follows Talend standard. 
Would take care later.


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/173/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/173/)

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedAvroRegistry.java
```diff
@@ -129,12 +129,16 @@ public String convertToDatum(T value) {
...
         private final Schema.Field field;
+        private final Character thousandsSepChar;
+        private final Character decimalSepChar;
 
         protected boolean isDecode;
 
         NumberConverter(Schema.Field field) {
             this.field = field;
             this.isDecode = Boolean.valueOf(field.getProp(ComponentConstants.NUMBER_DECODE));
+            this.thousandsSepChar = ParserUtils.parseToCharacter(field.getProp(ComponentConstants.THOUSANDS_SEPARATOR));
```
gonchar-ivan: why don't use separator value from properties?

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedAvroRegistry.java
```diff
@@ -185,6 +180,15 @@ public String convertToDatum(T value) {
             }
         }
 
+        /**
+         * Transform number string with thousandsSepChar and decimalSepChar
+         */
+        protected String transformNumberString(String value) {
+            if (thousandsSepChar != null && decimalSepChar != null) {
+                return ParserUtils.transformNumberString(value, thousandsSepChar, decimalSepChar);
```
gonchar-ivan: null checking could be moved inside transformNumberString() method, so you don't need transformNumberString() method in component

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedReader.java
```diff
@@ -80,21 +79,38 @@ protected void setupDynamicSchema() throws IOException {
...
         }
         return factory;
     }
 
-    /**
-     * Add decode property for field base on the decode table
-     */
-    protected void setupDecodeProps(Schema schema, TFileInputDelimitedProperties properties) {
-        if (schema != null && properties != null && properties.enableDecode.getValue()) {
+    protected void setupSchemaProps(Schema schema, TFileInputDelimitedProperties properties) {
```
gonchar-ivan: I think decode props shouldn't be added to schema

gonchar-ivan:
and, of course, commit messages should be changed according new format

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedAvroRegistry.java
```diff
@@ -171,6 +171,12 @@ public String convertToDatum(T value) {
                     } else {
                         return ((BigDecimal) value).toPlainString();
                     }
+                } else if (AvroUtils.isSameType(AvroUtils._decimal(), AvroUtils.unwrapIfNullable(field.schema()))) {
+                    String precision = field.getProp(SchemaConstants.TALEND_COLUMN_PRECISION);
+                    if (precision != null) {
```
jzhao-talend: In talend metadate schema "precision" can be null.

components-filedelimited/src/test/java/org/talend/components/filedelimited/runtime/FileDelimitedWriterTestIT.java
```diff
@@ -63,6 +67,31 @@ public void testOutputCsvStream() throws Throwable {
         testOutputCSV(true);
     }
 
+    @Test
+    public void testWriteDecimal() throws Throwable {
+        String resources = getResourceFolder();
```
jzhao-talend: sure. it can be improved

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedAvroRegistry.java
jzhao-talend: yes. you are right. transformNumberString can be removed

components-common/src/main/java/org/talend/components/common/ComponentConstants.java
```diff
@@ -5,12 +5,30 @@
  */
 public class ComponentConstants {
 
+    /**
+     * Use this property to specify number format thousands separator
+     */
```
jzhao-talend: Yes. They are used to convert to datum and to string .
Would be consider to replace with properties.

components-common/src/main/java/org/talend/components/common/ComponentConstants.java
```diff
@@ -31,4 +31,9 @@
      */
     public final static String CHECK_DATE = "check.date";
 
+    /**
+     * Specify whether or not decode the number string (decimal, hexadecimal, and octal numbers)
+     */
+    public final static String NUMBER_DECODE = "number.decode";
```
jzhao-talend: - provide at least javadoc comments
would add move javadoc
- You shouldn't add these constants/properties to Schema! 
would be replaced to use properties directly
- I propose you to improve ParserUtils class. For example, improve parseToInteger() method, add thousandSeparator parameter there etc.
yes. can be improve

components-common/src/main/java/org/talend/components/common/runtime/FastDateParser.java
```diff
@@ -15,7 +15,7 @@
 import java.util.Locale;
 
 /*
- * This is a cpty of routines.system.FastDateParser
+ * This is a copy of routines.system.FastDateParser
```
jzhao-talend: Not sure this currently. I didn't have a detail investigate. I think we can create a jira to investigate it later.

components-common/src/main/java/org/talend/components/common/runtime/ParserUtils.java
```diff
@@ -0,0 +1,219 @@
...
+/*
+ * This is a refer to routines.system.ParserUtils
+ */
+public class ParserUtils {
+
+    /**
+     * the source should be a string wrapped in chars[ ] which stands for it is a collection
+     * 
+     * @param listStr
+     * @param fieldSep
```
jzhao-talend: When I check the usage in component. I found it reuse the field separator in component setting.
Because of "field separator" have no limit in our component current.  I think that's why we didn't set the limitation before.

components-common/src/main/java/org/talend/components/common/runtime/ParserUtils.java
```diff
@@ -0,0 +1,219 @@
...
+     * @return
+     */
+    public static List<String> parseToList(final String listStr, String fieldSep) {
```
jzhao-talend: Thanks for your good suggestion.

components-common/src/main/java/org/talend/components/common/runtime/ParserUtils.java
```diff
@@ -0,0 +1,219 @@
...
+            }
+            result.append(strList.get(i));
+        }
+        result.append("]");
+
+        return result.toString();
+    }
+
+    /**
+     * Parse String type to character
```
jzhao-talend: agree with you

components-common/src/test/java/org/talend/components/common/runtime/ParserUtilsTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+public class ParserUtilsTest {
+
+    private transient static final Logger LOG = LoggerFactory.getLogger(FileRuntimeHelperTest.class);
+
+    @Test
+    public void testParseToList() throws Exception {
+        assertNull(ParserUtils.parseToList(null, ","));
+        // Test for list string not surrounded by "[ ]"
+        List<String> list1 = ParserUtils.parseToList("e1,e2,e3,e4", ",");
```
jzhao-talend: agree with you

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileSourceOrSink.java
```diff
@@ -135,7 +136,14 @@ public static Schema getDynamicSchema(String[] columnsName, String schemaName, S
                     }
                 }
             }
-            return Schema.createRecord(schemaName, null, null, false, fields);
+            Schema schema = Schema.createRecord(schemaName, null, null, false, fields);
+            if (designSchema.getProp(ComponentConstants.DIE_ON_ERROR) != null) {
```
jzhao-talend: would replace them to use properties.

components-filedelimited/src/test/java/org/talend/components/filedelimited/runtime/FileDelimitedReaderTestIT.java
```diff
@@ -28,15 +29,18 @@
...
..."di.table.label\":\"TestByte\",\"di.column.relatedEntity\":\"\"},{\"name\":\"TestShort\",\"type\":[{\"type\":\"int\",\"java-class\":\"java.lang.Short\"},\"null\"],\"di.table.comment\":\"\",\"talend.field.dbColumnName\":\"TestShort\",\"di.column.talendType\":\"id_Short\",\"di.column.isNullable\":\"true\",\"talend.field.pattern\":\"dd-MM-yyyy\",\"di.column.relationshipType\":\"\",\"di.table.label\":\"TestShort\",\"di.column.relatedEntity\":\"\"},{\"name\":\"TestInteger\",\"type\":[\"int\",\"null\"],\"di.table.comment\":\"\",\"talend.field.dbColumnName\":\"TestInteger\",\"di.column.talendType\":\"id_Integer\",\"di.column.isNullable\":\"true\",\"talend.field.pattern\":\"\",\"di.column.relationshipType\":\"\",\"di.table.label\":\"TestInteger\",\"di.column.relatedEntity\":\"\"},{\"name\":\"TestLong\",\"type\":[\"long\",\"null\"],\"di.table.comment\":\"\",\"talend.field.dbColumnName\":\"TestLong\",\"di.column.talendType\":\"id_Long\",\"di.column.isNullable\":\"true\",\"talend.field.pattern\":\"\",\"di.column.relationshipType\":\"\",\"di.table.label\":\"TestLong\",\"di.column.relatedEntity\":\"\"}],\"di.table.name\":\"tFileInputDelimited_1\",\"di.table.label\":\"tFileInputDelimited_1\"}");
```
jzhao-talend: you are right. This is copied from job generated code.
It's too di special. I would replaced it with a common and clear one.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedReader.java
jzhao-talend: would replace them to use properties.

jzhao-talend:
Need review and merge https://github.com/Talend/daikon/pull/69 firstly.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/249/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/249/)

jzhao-talend:
About this commit https://github.com/Talend/components/pull/260/commits/8a39ff7d03376415e965af7ab415fb1563815f83

- [x] Fix the wrong usage of convert
- [x]  Remove the ComponentConstants which set in runtime schem for AvroConvert and replace it with Runtime properties

The issue would be fix in other jira:

- Improve the ParseUtil 
- The unit test about get dynamic schema which included some special di code. 
Need to investigate modify it or just remove them from tcomp test.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedAvroRegistry.java
```diff
@@ -130,17 +139,28 @@ public String convertToDatum(T value) {
 
         private final Schema.Field field;
 
-        private final Character thousandsSepChar;
+        protected Character thousandsSepChar;
+
+        protected Character decimalSepChar;
 
-        private final Character decimalSepChar;
+        protected final FileDelimitedProperties properties;
```
gonchar-ivan: No need to store properties as a field in NumberConverter. Seems, you just retrieve values of decimalSepChar and thousandsSepChar from it

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/FileDelimitedAvroRegistry.java
```diff
@@ -130,17 +139,28 @@ public String convertToDatum(T value) {
...
-        protected boolean isDecode;
+        private List<Boolean> decodeList;
 
-        NumberConverter(Schema.Field field) {
+        NumberConverter(Schema.Field field, FileDelimitedProperties properties) {
             this.field = field;
-            this.isDecode = Boolean.valueOf(field.getProp(ComponentConstants.NUMBER_DECODE));
-            this.thousandsSepChar = ParserUtils.parseToCharacter(field.getProp(ComponentConstants.THOUSANDS_SEPARATOR));
-            this.decimalSepChar = ParserUtils.parseToCharacter(field.getProp(ComponentConstants.DECIMAL_SEPARATOR));
+            this.properties = properties;
```
gonchar-ivan: No need to store properties

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/DelimitedAdaptorFactory.java
```diff
@@ -62,13 +74,24 @@ public Object get(int index) {
             // Lazy initialization of the cached converter objects.
             Object value = null;
             if (trimValues == null) {
+                if (properties == null) {
```
gonchar-ivan: hmm... I have just realized the idea of this DelimitedIndexedRecord.
It stores field values as String[] and uses "lazy initialization of converter.
Interesting.

For now, such solution works.
But imagine situation in distributed environment:
FileDelimitedInput component runs on one computer and next component runs on other computer.
So, environment gets IndexedRecords from component, serialize them and send to another computer via network.
To deserialize and use them on another computer that computer should have installed(loaded by ClassLoader) FileDelimited converters, but it may not have it (with avro it is possible to deserialize data even if don't have exact java type for it)

So, I don't understand the reason for laze initialization here

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/DelimitedAdaptorFactory.java
```diff
@@ -53,44 +65,48 @@ public void put(int index, Object json) {
...
 
         @Override
         public Object get(int index) {
             // Lazy initialization of the cached converter objects.
             Object value = null;
-            if (names == null) {
-                names = new String[getSchema().getFields().size()];
-                fieldConverter = new AvroConverter[names.length];
-                for (int j = 0; j < names.length; j++) {
+            if (trimValues == null) {
```
gonchar-ivan: hmm... I have just realized the idea of this DelimitedIndexedRecord.
It stores field values as String[] and uses "lazy initialization of converter.
Interesting.

For now, such solution works.
But imagine situation in distributed environment:
FileDelimitedInput component runs on one computer and next component runs on other computer.
So, environment gets IndexedRecords from component, serialize them and send to another computer via network.
To deserialize and use them on another computer that computer should have installed(loaded by ClassLoader) FileDelimited converters, but it may not have it (with avro it is possible to deserialize data even if don't have exact java type for it)

So, I don't understand the reason for laze initialization here

jzhao-talend: Seems yes. we didn't consider this kind of solution.
The first usage of  "lazy initialization of converter“ is in salesforce. 
https://github.com/Talend/components/commit/08715d42804e0bf1b6b229eeb3d3dfdf4ddf8fe4
So maybe we can discuss with Ryan later. I think you can create a jira for this.

PR #267, opened 2016-10-19
fix (TDI-37564): DND problem of filedelimited components

jzhao-talend:

1. Can't use wizards to drag & drop to create components
2. "Property Type" is missing in tFileOutputDeclimited

**What is the new behavior?**
Refactor the properties and use FileDelimitedProperties as wizard properties to fix above issue

- [x] No

**Other information**:
Only add the test that we can add in components project
About the drag and drop it should be test in studio.


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/269/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/269/)

components-filedelimited/src/main/java/org/talend/components/filedelimited/wizard/FileDelimitedWizardProperties.java
```diff
@@ -1,135 +0,0 @@
-package org.talend.components.filedelimited.wizard;
```
gonchar-ivan: Yes, it was incorrect to create separate properties for wizard. 
For wizard you need create separate form. Ok, seems I understood the reason of a problem

components-filedelimited/src/main/java/org/talend/components/filedelimited/FileDelimitedProperties.java
```diff
@@ -124,4 +185,68 @@ public void setSchemaListener(ISchemaListener schemaListener) {
...
+    public ValidationResult afterFormFinishWizard(Repository<Properties> repo) throws Exception {
+        // Clear the preview data when form is finished
+        previewTable.setValue("{\"data\":[]}");
+        String connRepLocation = repo.storeProperties(this, this.getName(), repositoryLocation, null);
+        repo.storeProperties(this, this.name.getValue(), connRepLocation, "main.schema");
+        return ValidationResult.OK;
+    }
+
+    public FileDelimitedProperties setRepositoryLocation(String location) {
+        repositoryLocation = location;
```
jzhao-talend: Because of I can't debug the code of wizard. I just know that when wizard init would call this method to pass the wizard repositoryLocation to it.

components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/DelimitedReader.java
```diff
@@ -21,7 +22,7 @@
 
     private FileInputDelimited fid;
 
-    public DelimitedReader(RuntimeContainer container, BoundedSource source, TFileInputDelimitedProperties properties) {
+    public DelimitedReader(RuntimeContainer container, BoundedSource source, FileDelimitedProperties properties) {
```
jzhao-talend: yes. the this part of code need to improve.
I would refactor the code after 630.

components-filedelimited/src/main/java/org/talend/components/filedelimited/FileDelimitedProperties.java
```diff
@@ -53,21 +68,45 @@ public void afterSchema() {
...
         fieldSeparator.setValue(";");
         escapeChar.setValue("\\\"");
         textEnclosure.setValue("\\\"");
         thousandsSeparator.setValue(",");
         decimalSeparator.setValue(".");
-        encoding.encodingType.setPossibleValues(encoding.getDefaultEncodings());
+        header.setValue(0);
+        footer.setValue(0);
+        removeEmptyRow.setValue(true);
+        encoding.encodingType.setPossibleValues(Arrays.asList(ENCODING_TYPE_UTF_8, "UTF-16", "UTF-16LE", "UTF-16BE", "UTF-7",
```
jzhao-talend: >> Could this list of charsets be an enum inside EncodingTypeProperties?
Yes. we have a jira to do it. Because of time limitation. I didn't do it here.

PR #269, opened 2016-10-19
components-filedelimited: Fix initialization code for avro converters.

RyanSkraba:

The tFileInputDelimited re-initializes avro convert objects per-record when they can be reused for the entire job.

**What is the new behavior?**

The tFileInputDelimited reuses these objects for the entire job.

- [ ] Yes
- [X] No

**Other information**:

Tests already exist.  I didn't find a JIRA for the performance issue.

### DIFF
--- a/components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/DelimitedAdaptorFactory.java
+++ b/components-filedelimited/src/main/java/org/talend/components/filedelimited/runtime/DelimitedAdaptorFactory.java
@@ -19,6 +19,10 @@ public class DelimitedAdaptorFactory implements IndexedRecordConverter<String[],
 
     FileDelimitedProperties properties;
 
+    private boolean trimValues[];
+
+    private AvroConverter[] fieldConverter = null;
+
     @Override
     public Class<String[]> getDatumClass() {
         return String[].class;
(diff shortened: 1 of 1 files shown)


RyanSkraba:
# Previous 6.3 tFileInput timing (consuming 1 million ten-column rows from a 0.5GB file)

```
/tmp/profiler/before/file02_consume_1M_rows$ for i in {1..10}; do echo $i -----; time sh file02_consume_1M_rows_run.sh ; done
1 -----

real	0m33.890s
user	0m40.376s
sys	0m0.316s
2 -----

real	0m30.979s
user	0m37.324s
sys	0m0.272s
3 -----

real	0m30.978s
user	0m38.252s
sys	0m0.264s
4 -----

real	0m31.131s
user	0m37.672s
sys	0m0.276s
5 -----

real	0m31.522s
user	0m37.948s
sys	0m0.312s
6 -----

real	0m35.160s
user	0m45.468s
sys	0m0.240s
7 -----

real	0m33.923s
user	0m41.048s
sys	0m0.300s
8 -----

real	0m30.634s
user	0m37.360s
sys	0m0.324s
9 -----

real	0m31.486s
user	0m38.356s
sys	0m0.212s
10 -----

real	0m30.735s
user	0m36.288s
sys	0m0.276s
```

# After this change

```
/tmp/profiler/after/file02_consume_1M_rows$ for i in {1..10}; do echo $i -----; time sh file02_consume_1M_rows_run.sh ; done
1 -----

real	0m6.346s
user	0m7.464s
sys	0m0.176s
2 -----

real	0m5.858s
user	0m7.224s
sys	0m0.184s
3 -----

real	0m6.516s
user	0m7.700s
sys	0m0.184s
4 -----

real	0m6.397s
user	0m7.724s
sys	0m0.172s
5 -----

real	0m5.650s
user	0m6.992s
sys	0m0.148s
6 -----

real	0m5.265s
user	0m6.652s
sys	0m0.160s
7 -----

real	0m6.338s
user	0m7.516s
sys	0m0.192s
8 -----

real	0m6.286s
user	0m7.504s
sys	0m0.196s
9 -----

real	0m6.305s
user	0m7.540s
sys	0m0.124s
10 -----

real	0m5.474s
user	0m6.748s
sys	0m0.240s
```

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/282/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/282/)

francisu:
Looks good, merge as you like!

gonchar-ivan:
Also this PR LGTM. 
However, I have a question to @RyanSkraba 

This component stores values as String[] inside IndexedRecord and converts values to real types only during extraction ( inside IndexedRecord.get(int i) method ).
What do you think: is it ok to store values that way. Or is it better to story values already converted to their actual types.

I just imagine following situation:
We want to serialize IndexedRecord (using avro serialization) just after FileInputDelimited component. So, IndexedRecord.get wasn't called. 
After deserialization such IndexedRecord on other computer we are trying to call get() method. In that moment corresponding converters are required, but they could be not available, because other computer has no Converter class files.
Is such situation reasonable or it is improbable?

RyanSkraba:
Hello!

I think it would be better to store values already converted to their actual types -- but this doesn't make any difference in the Studio because they will only be converted exactly once (when they are copied out using those DiEnforcers!).

In another environment, it's possible that we'll access some values more than once, so we should convert them more only once.

It's neat on a cluster, however -- when it happens that data is passed from one node to another, it will be serialized to Avro (binary), passed over the net, then deserialized back into an Avro IndexedRecord.  The second IndexedRecord loses the "implementation" of the first and should just contain the Avro values.

... There are still some details to work out ...

gonchar-ivan:
Ryan, thanks)

PR #270, opened 2016-10-20
TDI-37658: avoid the NPE when use tsalesforcebulkexec and

wwang-talend:
https://jira.talendforge.org/browse/TDI-37658


francisu:
unit test please.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/298/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/298/)

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforcebulkexec/TSalesforceBulkExecDefinition.java
```diff
@@ -62,11 +62,7 @@ public boolean isConditionalInputs() {
 
     @Override
     public RuntimeInfo getRuntimeInfo(Properties properties, ConnectorTopology componentType) {
-        if (componentType == ConnectorTopology.OUTGOING) {
-            return getCommonRuntimeInfo(this.getClass().getClassLoader(), SalesforceSource.class);
-        } else {
-            return null;
-        }
+        return getCommonRuntimeInfo(this.getClass().getClassLoader(), SalesforceSource.class);
```
wwang-talend: yes

gonchar-ivan: As for caller should handle null value:
RuntimeInfo runtimeInfo = definition.getRuntimeInfo(properties, componentType);
if (runtimeInfo == null) {
   throw new Exception("Please, use correct connection"); 
}

So, for my opinion there is a bug in Studio side. What do you think about it?

wwang-talend: agree, but this is some risk for that as the getRuntimeInfo method will be called by many places, not only the job runtime. This is a quick fix, we can do it later.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/302/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/302/)

components-salesforce/src/main/java/org/talend/components/salesforce/tsalesforcebulkexec/TSalesforceBulkExecDefinition.java
gonchar-ivan: Even javadoc says that such method may return null
https://github.com/Talend/components/blob/master/components-api/src/main/java/org/talend/components/api/component/ComponentDefinition.java#L117-L128

;-)

wwang-talend:
Hi Ivan, how about the last commit? it follow the document.

gonchar-ivan:
ok for me

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/305/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/305/)

PR #280, opened 2016-10-28
feat(TCOMP-256): Add relationship between Datastore/Dataset/IO properties

bchen-talend:

The datastore and dataset properties are related by composition, but there is currently no way to navigate between the two.

**What is the new behavior?**

Add methods to the respective interfaces to get and set the datastore properties from a dataset and a dataset properties from an IOComponent.

- [ ] Yes
- [x] No

There is no breaking change because no implementations of Datastore and DatasetProperties exist yet on master.  Working branches will need to be updated for components in progress.

**Other information**:


build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/417/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/417/)

components-common/src/main/java/org/talend/components/common/io/IOProperties.java
```diff
@@ -0,0 +1,21 @@
+package org.talend.components.common.io;
+
+import org.talend.components.common.FixedConnectorsComponentProperties;
+import org.talend.components.common.dataset.DatasetProperties;
+
+public abstract class IOProperties<DS extends DatasetProperties> extends FixedConnectorsComponentProperties {
+
+    /**
+     * FixedSchemaComponentProperties constructor comment.
```
RyanSkraba: Bad comment :)

components-common/src/main/java/org/talend/components/common/datastore/runtime/DatastoreRuntime.java
```diff
@@ -17,11 +17,11 @@
 import org.talend.components.common.datastore.DatastoreProperties;
 import org.talend.daikon.properties.ValidationResult;
 
-public interface DatastoreRuntime extends RuntimableRuntime<DatastoreProperties> {
+public interface DatastoreRuntime<T extends DatastoreProperties> extends RuntimableRuntime<T> {
```
RyanSkraba: Good idea!  Can you name the generic type DatastorePropT like above?

components-common/src/main/java/org/talend/components/common/io/IOProperties.java
```diff
@@ -0,0 +1,21 @@
+package org.talend.components.common.io;
+
+import org.talend.components.common.FixedConnectorsComponentProperties;
+import org.talend.components.common.dataset.DatasetProperties;
+
+public abstract class IOProperties<DS extends DatasetProperties> extends FixedConnectorsComponentProperties {
```
RyanSkraba: Can you rename the generic type as DatasetPropT to be consistent as above?

RyanSkraba:
When the build fails because of another component, can you just comment to verify that it succeeds locally?  I don't know what else we can do when this happens...

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/462/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/462/)

bchen-talend:
Thanks Ryan, I have changed as you said, and again it failed by some other component.

sgandon:
There is explanation of this PR in the description.
Please fill *What is the current behavior? (You can also link to an open issue here)*, *What is the new behavior?*

RyanSkraba:
Hello!  I added more information to the description of this PR as I understood it -- don't hesitate to correct my interpretation.

This is a good idea because we will need to navigate between IOComponent/Dataset/Datastore properties -- we expect all the input/output components to have this hierarchy via property composition.

components-common/src/main/java/org/talend/components/common/io/IOProperties.java
```diff
@@ -0,0 +1,21 @@
+package org.talend.components.common.io;
+
+import org.talend.components.common.FixedConnectorsComponentProperties;
+import org.talend.components.common.dataset.DatasetProperties;
+
+public abstract class IOProperties<DatasetPropT extends DatasetProperties> extends FixedConnectorsComponentProperties {
```
RyanSkraba: It just occurred to me -- it might be better to have this as an interface with the getDatasetProperties() and setDatasetProperties() method.  Input/Output components might not always want to extend FixedConnectorsComponentProperties, but should always implement this interface!

What do you think?

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/530/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/530/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/531/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/531/)

components-common/src/main/java/org/talend/components/common/dataset/DatasetProperties.java
```diff
@@ -1,9 +1,14 @@
...
 
 /**
  * Placeholder for DatasetProperties.
  */
-public interface DatasetProperties extends Properties {
+public interface DatasetProperties<DatastorePropT extends DatastoreProperties> extends Properties {
+
+    public DatastorePropT getDatastoreProperties();
+
+    public void setDatastoreProperties(DatastorePropT datastoreProperties);
```
sgandon: I am wondering if we need those apis, and otherwise use the ComponentReferenceProperties (but of course generalising it to simple Properties not only ComponentProperties).
This would allow us to handle this as a reference to some other properties and nothing specific to Datasets/Datastores. Because we will also have other references too like security seetings and so on.

bchen-talend: There is some different usage for the current ComponentReferenceProperties, 
ComponentReferenceProperties only store id of the relate component id, then use this component id on runtime(globalMap.get), which is special concept for Studio.

But after look at the DatastoreDefinition.createDatasetProperties, I think I can remove get/setDatastoreProperties in DatasetProperties, and remove IOProperties totally.
I made this change before I know DatastoreDefinition.createDatasetProperties, and the initial idea is expose a set method for rest service to set relative properties.

But need to add DatasetDefinition.createComponentProperties(T extends DatsetProperties) for return Input/Output component properties.
@sgandon, @RyanSkraba what do you think?

RyanSkraba:
This looks good to me -- since it only affects TFD components, we can use it as a good starting point and make any adjustments if they are required.  We can come back to the issue of non-Datastore, Dataset referenced properties later.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/546/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/546/)

sgandon:
This is ok with me for now. I really believe we should use ReferenceProperties but w'ill do that later.
Please can you wait for the newbuild nexus server to come back online before merging.

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/551/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/551/)

PR #303, opened 2016-11-15
TDI-37812 Snowflake component tests are not properly isolated.

francisu:

Tests fail when run concurrently because they used the same schema.

**What is the new behavior?**

Tests work when run concurrently (tested 3 concurrent tests)

- [ ] Yes
- [X] No

**Other information**:

A couple of other small corrections. Also removed the running of the plain tests (by making it an abstract class), only the Spring and OSGi tests are being used.


components-api/src/test/java/org/talend/components/api/properties/ComponentPropertiesTest.java
```diff
@@ -75,6 +76,7 @@ public void init() {
     @Test
     public void testUpdateNestedProperties() throws IllegalAccessException {
         NestedNestedProperties nestedProperties = new NestedNestedProperties("bar");
+        System.out.println(SerializerDeserializer.toSerializedPersistent(nestedProperties));
         nestedProperties.value.setValue("XYZ");
```
francisu: Not really part of the change; but put in to show someone something. I think leaving it is harmless.

sgandon: why don't you use a slf4j Logger ?

sgandon:
LGTM

PR #305, opened 2016-11-16
TDI-37807: [tDataSetInput] the wizard does not fetch schema dataset with

wwang-talend:

https://jira.talendforge.org/browse/TDI-37807

**What is the new behavior?**
correct the name

- [ ] Yes
- [x] No

**Other information**:

special characters in column names
https://jira.talendforge.org/browse/TDI-37807


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/745/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/745/)

fhuaulme:
I find the method Character.isValidJavaIdentifier() (not actual name, but rather close) easier to read.

And your fix doesn't cover case where columns are {"éà", "+_"}, it would end up with 2 columns named "__" and Avro doesn't like this.

wwang-talend:
About the first advice, please see the spec of avro name : http://avro.apache.org/docs/current/spec.html#names
the changeset follow it totally, and i also try to find a easy readable method in java.lang.Character, no found.

About the second advice, agree it totally, i will fix it asap

wwang-talend:
About the first advice, please see the spec of avro name : http://avro.apache.org/docs/current/spec.html#names
the changeset follow it totally, and i also try to find a easy readable method in java.lang.Character, no found.

About the second advice, agree it totally, i will fix it asap.

Thanks.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/750/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/750/)

wwang-talend:
have fixed it

components-dataprep/src/test/java/org/talend/components/dataprep/util/NameUtilTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.talend.components.dataprep.connection.DataPrepField;
+import org.talend.components.dataprep.runtime.DataPrepAvroRegistry;
+import org.talend.daikon.avro.AvroRegistry;
+
+public class NameUtilTest {
+
+    @Test
+    public void testCorrect() {
```
gonchar-ivan: The idea of unit-test not only verify that there is no error, but also save time when you are looking a reason of a bug.

Normally, when you see that NameUtilTest.testCorrect() test-case fails it should mean that a reason is in NameUtil.correct() method.

But this is not correct for your test.
In your test-case you check NameUtil.correct() method indirectly. So, when test-case fails reason could be not only from NameUtil.correct(), but also in DataPrepAvroRegistry. We will need to spend additional time to find reason in case of test fail.

However, such test-case as you written also is useful.
It allows to see that everything work fine together. And this test-case is integration test-case.

We need both: unit-tests and ITs for components.

components-dataprep/src/main/java/org/talend/components/dataprep/util/NameUtil.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    private static String getUniqueName(String name, List<String> previousNames) {
+        boolean allIsDifferent = false;
+        int index = 0;
+        String currentName = name;
+        while (!allIsDifferent) {
+            allIsDifferent = true;
+
+            for (String each : previousNames) {
```
gonchar-ivan: UPD:

please, use HashSet instead of List
and HashSet.contains()

components-dataprep/src/main/java/org/talend/components/dataprep/runtime/DataPrepAvroRegistry.java
```diff
@@ -84,11 +85,15 @@ public String getFamilyName() {
      */
     private Schema inferSchemaDataPrepResult(DataPrepField[] in) {
         List<Schema.Field> fields = new ArrayList<>();
+        int index = 0;
+        List<String> existNames = new ArrayList<String>();
```
gonchar-ivan: Please, use HashSet instead of List

components-dataprep/src/main/java/org/talend/components/dataprep/util/NameUtil.java
```diff
@@ -0,0 +1,58 @@
+package org.talend.components.dataprep.util;
+
+import java.util.List;
+
+public class NameUtil {
+
+    public static String correct(String name, int nameIndex, List<String> previousNames) {
```
wwang-talend: agree it, will do it

components-dataprep/src/main/java/org/talend/components/dataprep/util/NameUtil.java
```diff
@@ -0,0 +1,58 @@
...
+        if (name == null || name.isEmpty() || previousNames == null) {
+            return name;
+        }
+
+        StringBuilder str = new StringBuilder();
+        int underLineCount = 0;
+        for (int i = 0; i < name.length(); i++) {
+            char c = name.charAt(i);
+            if (((c >= 'a') && (c <= 'z')) || ((c >= 'A') && (c <= 'Z')) || ((c >= '0') && (c <= '9') && (i != 0))) {
```
wwang-talend: i have a different idea about it, the pattern is heavy sometimes like this case,  i need more than one pattern for it in this method, also i need to consider the cache about it. My implement is easy to understand and not a complex one and maybe a better performance. If the pattern is better and more easy, i think i will use it as i think pattern is easy sometimes for reading, but at this place, i don't think so.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/869/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/869/)

wwang-talend:
have fixed it

components-dataprep/src/main/java/org/talend/components/dataprep/util/NameUtil.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        return getUniqueName(result, previousNames);
+    }
+
+    private static String getUniqueName(String name, Set<String> previousNames) {
+        boolean allIsDifferent = false;
+        int index = 0;
+        String currentName = name;
+        while (!allIsDifferent) {
```
gonchar-ivan: I suggest:

```java
while(previousNames.contains(currentName)) {
    currentName = currentName + (++index); 
}
```

Just FYI, when you use "+" (concat) in a loop new instance of StringBuilder is created each iteration (if it wasn't optimized in recent Java version). I hope there won't be a lot of iterations here, so we can leave as is.

components-dataprep/src/main/java/org/talend/components/dataprep/util/NameUtil.java
gonchar-ivan: ok

PR #311, opened 2016-11-18
bug(serialization): Error example.

pbailly:

We currently cannot serialize a property with a schema. (See the test TFakeLogRuntimeBuggyTest)
We are forced to serialize by hand the properties in order to be able to use the component in beam. (See TFakeLogRuntimeTest)

It would be really useful set up an automatic serialization of any Property.

**What is the expected behavior?**

- [ ] Yes
- [ ] No

**Other information**:


sgandon:
The reason for the SchemaProperties to fail on serialization is that the object is not serializable.
Chen Bin will make a PR to make it serializable.
I'll close this PR as it was just for discussion.

PR #313, opened 2016-11-21
Feature/tcomp 262 datastore rest endpoint suite

sgandon:

endpoints for the web services do not match the requirements

**What is the new behavior?**
Now the endpoints match the requirements
We had to remove enunciate and therefore no more web API docs are generated, we need to fix this.
There are some minor fixes on the components themselves too.

- [x] No


services/components-api-service-rest/src/main/java/org/talend/components/service/rest/RuntimesController.java
```diff
@@ -0,0 +1,45 @@
...
+    /** Validate connection to datastore. Should it be GET method? **/
+    // P2
+    @RequestMapping(value = "{definitionName}/schema", method = POST)
+    void getDatasetSchema(@PathVariable("definitionName") String definitionName);
+
+    /** Get dataset content. Should it be GET method? **/
+    // P2
+    @RequestMapping(value = "{definitionName}/data", method = POST)
+    void getDatasetSchema(@PathVariable("definitionName") String definitionName,
+                          @RequestParam(value = "from", defaultValue = "1000") Integer from,
```
RyanSkraba: Quick comment here -- from the last discussion that we had, "from" is not supported.

services/components-api-service-rest/src/main/java/org/talend/components/service/rest/DefinitionType.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (C) 2006-2015 Talend Inc. - www.talend.com
```
RyanSkraba: 2016 everywhere :)  There's also some source files with missing headers.

sgandon:
I have created a Jira for the missing and erroneous java headers
https://jira.talendforge.org/browse/TCOMP-357

PR #321, opened 2016-11-23
fix(TDI-37870): replace json-simple with json-io in Splunk component

chmyga:

Splunk component used json-simple library

**What is the new behavior?**
Splunk component will use json-io library. All components will use the same version of json-io, set in components-parent pom.xml dependency management block

- [ ] Yes
- [x] No

**Other information**:

* Add json-io to dependency management block of components-parent to use
* the same version in all components
* Replace json-simple with json-io in Splunk component


components-parent/pom.xml
```diff
@@ -98,6 +98,11 @@
 				<artifactId>avro</artifactId>
 				<version>1.8.0</version>
 			</dependency>
+			<dependency>
```
sgandon: We have already a json-io in the daikon depenencies so you should not need to add it here.

components-splunk/pom.xml
```diff
@@ -44,17 +44,10 @@
...
-			<artifactId>json-simple</artifactId>
-			<version>1.1.1</version>
-			<exclusions>
-			     <exclusion>
-			         <groupId>junit</groupId>
-			         <artifactId>junit</artifactId>
-			     </exclusion>
-			</exclusions>
+			<groupId>com.cedarsoftware</groupId>
+			<artifactId>json-io</artifactId>
```
sgandon: same, you should have acces to json-io though transitive depenencies of daikon.

components-splunk/src/main/java/org/talend/components/splunk/objects/SplunkJSONEvent.java
```diff
@@ -12,39 +12,38 @@
...
+public class SplunkJSONEvent extends HashMap<String, Object> {
 
     private static final long serialVersionUID = 3205801702373951474L;
 
     public SplunkJSONEvent() {
     }
 
     public void setEventObjects(Map<String, Object> eventObjects) {
-        JSONObject jsonEventObject = new JSONObject(eventObjects);
+        Map<String, Object> jsonEventObject = new HashMap<>(eventObjects);
```
sgandon: why did not you use the JsonObject provided by Json-io ?

components-splunk/pom.xml
gonchar-ivan: I have following thoughts about it:

I often use pom.xml to check what libraries are directly used by module. When json-io comes from daikon for some other component (e.g. Jira), it is not called directly in Jira. But here json-io is called directly.

So, it could be placed here for informational purposes.

Imagine situation:
Assume, we want to get rid of json-io library (or some other, which comes from daikon). 
We know that some of components use it and we need to change some components.
But we don't know what exact components use it.
If this is library is present as dependency in those components, we can find those components via search upon pom.xml files. 
If it is not present in pom.xml we need to rebuild full repository to check where compile error will occur.

gonchar-ivan: Another thing - we want to align libraries version, which are used in daikon, components API and components themselves.
We can use dependencyManagement blocks for it. But we have several parent poms:
component-parent, daikon-parent (soon component-api-parent)
So, there will be several places to look about library version to use.

What do you think about following idea:
daikon-parent pom will have dependencyManagement block for its dependencies. Obviously that components-api-parent and components-parent should use the same versions. 
Could we import dependencyManagement from daikon-parent to components-parent?
(The same way as spring-boot suggest - in scope import)

sgandon:
I agree with your comments, you are right and should be adding the required dependencies explicitly like json-io.
On the version side, I think we could add daikon in the <dependencyManagement> with the scope <import>. Do you think you can do it for this PR?

chmyga:
@sgandon Maybe we can do it in scope of separate issue, as this would affect daikon too, not only the components repo?

sgandon:
ok.

gonchar-ivan:
created https://jira.talendforge.org/browse/TCOMP-374 to do what we discussed

PR #322, opened 2016-11-23
Bugfix/tdi 37873

francisu:

https://jira.talendforge.org/browse/TDI-37873

**What is the new behavior?**

Component works in the studio

- [ ] Yes
- [X] No

**Other information**:


build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/911/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/911/)

sgandon:
checking the tests issues.

sgandon:
OSGI tests are failing because you removed the `,com.snowflake*` from the export package command.
I think this cause not arm to export those API and will make the tests work. I also wonder if you are not required to export them to be used by the jdbc APIs.

francisu:
OK, I fixed the export; The problem is the export causes the classes to be included in files inside of the Jar files; but the studio requires the actual Jar file. Both are now being included (at a cost of twice the size -- which is unfortunate, but at least everything will work).

PR #325, opened 2016-11-24
Bugfix/tcomp 357 copyright

bchen-talend:

there is no copyright checker/format

**What is the new behavior?**
add copyright checker and format on maven build stage
and correct some class by this maven plugin

- [ ] Yes
- [x] No

**Other information**:


bchen-talend:
Hi @sgandon,
I added a maven license plugin which almost works,
now it can add copyright header auto if the class do not have when maven build, and let the developer commit the change.
or I can remove the format goal and left the check only, then maven build will failed if some class missing copyright header.

And you will find I have to add this plugin in the project which do not dependent on component-parent, such as **components-adapter-beam**.
There is another project **components-api-service-rest** also, but it dependent on daikon-parent, so I do not modify it this time, and suppose we may modify daikon-parent.

And there is also a build failed problem I don't know how to fix it properly
```
[ERROR] Failed to execute goal org.commonjava.maven.plugins:directory-maven-plugin:0.1:directory-of (directories) on project file-input: Cannot find directory for pro
ject: org.talend.components:components-parent 
```
directory-maven-plugin is added for get the absolutely path of license template which in components-parent folder.
but cause by file-input is fake.

pteyssier:
Work in progress since nearly one year, should not be in a PR

PR #336, opened 2016-11-28
fix (TDI-37208): NPE when tsalseforceOutput action is upsert

jzhao-talend:

When schema include field which not exist in module. and it is used in upsert relation table.
Would get NPE.

**What is the new behavior?**
Avoid the NPE

- [ ] Yes
- [x] No

**Other information**:
Upsert relation table can't add junit with current salesforce test account.


jzhao-talend:
Merge PR test is not reported.
Need to wait https://jira.talendforge.org/browse/DEVOPS-1413

gonchar-ivan:
Maybe you may try to rerun Jenkins. I know 2 ways to do it:
1. Rebuild button in Jenkins build page
https://ci-common.datapwn.com/job/components_pull-request_master/1069/
2. Update PR:
2.1. git commit --amend (change something. e.g. add space in commit message)
2.2. git push -f origin [branch]
or add one more fake commint

jzhao-talend:
@gonchar-ivan 
Thanks. Seems solution 1 is works. And the build is block by another problem.

jzhao-talend:
@sgandon 
As what I said in pull request description, I didn't find a way to use salesforce default standard fields to test upsert relation table.
Should I add some unit test and ignore  them during build. Just test them manually?
Or can I get the test salesforce account to create some test module for upsert relation ?

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/1105/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/1105/)

jzhao-talend:
I have add the unit test but with "@ignore"
If we want remove "@ignore", we need to add 2 custom module to the test salesforce account:
1. Object Name:"TestLookupModule"(API Name: "TestLookupModule__c")
Create External ID field for "TestLookupModule":
**Field Type:** "TEXT"
**Field Name:** "ExternalID"(API Name: "ExternalID__c")
**_Note: Check "Set this field as the unique record identifier from an external system", when created the field_**
2. Object Name:"TestModule"(API Name: "TestModule__c")
a) **Create External ID field for "TestLookupModule":**
**Field Type:** "TEXT"
**Field Name:** "ExternalID"(API Name: "ExternalID__c")
**_Note: Check "Set this field as the unique record identifier from an external system", when created the field_**
b) **Created lookup field for "TestModule"**
**Field Type:** "Lookup Relationship"
Choose the related object: "TestLookupModule"
Field Name: "TestLookupModule"(API Name: "TestLookupModule__c")

build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1148/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1148/)

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/1152/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/1152/)

jzhao-talend:
@sgandon

components-salesforce/src/main/java/org/talend/components/salesforce/runtime/SalesforceWriter.java
```diff
@@ -208,27 +207,38 @@ private SObject createSObjectForUpsert(IndexedRecord input) {
         nullValueFields.clear();
         for (Schema.Field f : input.getSchema().getFields()) {
             Object value = input.get(f.pos());
-            Schema.Field se = moduleSchema.getField(f.name());
-            if (value != null && !"".equals(value.toString()) && se != null) {
+            Schema.Field se = mainSchema.getField(f.name());
```
wwang-talend: but strange thing to me, for other action, now we use runtime schema to filter the input schema, i think we should make the filter be the same, use the design schema. We can adjust it later if it's dangerous, now it's ok to me.

wwang-talend: for example, tfixflowinput(c1,c2,c3)==>tsalesforceoutput(c1,c2)(suppose it's valid, not sure), the salesforce module schema is like this (c1,c2,c3,c4,c5,c6), then the c3 value from the input record of tfixflowinput will be inserted to salesforce, i think it's not expected. But there is no relation with current fix, we can consider it later.

wwang-talend:
in my opinion, we better to use the same word for the avro design schema(which is stored and load when use the component) and avro runtime schema(at runtime, we fetch it for something like dynamic or use it like here), and talend schema, and i think main/output/reject schema should alreadys belong to avro design schema.

PR #340, opened 2016-11-28
Bugfix/6.3/tdi 36867 jira error message

gonchar-ivan:

Jira components doesn't provide clear error message in case of exception

**What is the new behavior?**

Jira components provide error message from Jira server

- [ ] Yes
- [x] No

**Other information**:

Related to https://jira.talendforge.org/browse/TDI-36867


build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/1097/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/1097/)

components-jira/src/main/java/org/talend/components/jira/connection/Rest.java
```diff
@@ -143,7 +150,16 @@ public String get(String resource, Map<String, Object> parameters) throws IOExce
                 get.addHeader(header);
             }
             executor.clearCookies();
-            return executor.execute(get).returnContent().asString();
+            Response response = executor.execute(get);
+            HttpResponse httpResponse = response.returnResponse();
+            HttpEntity entity = httpResponse.getEntity();
+            String entityBody = "";
```
wwang-talend: i guess this method will not fetch the huge data, so we can write it like this.

components-jira/src/main/java/org/talend/components/jira/connection/Rest.java
```diff
@@ -194,35 +219,53 @@ public int delete(String resource, Map<String, Object> parameters) throws IOExce
...
         }
         executor.clearCookies();
-        return executor.execute(put).returnResponse().getStatusLine().getStatusCode();
+        Response response = executor.execute(put);
+        HttpResponse httpResponse = response.returnResponse();
+        StatusLine statusLine = httpResponse.getStatusLine();
+        int statusCode = statusLine.getStatusCode();
+        HttpEntity entity = httpResponse.getEntity();
+        String entityBody = "";
+        if (entity != null && statusCode != SC_UNAUTHORIZED) {
```
wwang-talend: i see statusCode SC_UNAUTHORIZED check for all request method except the get method, any reason? if the get method also need it, i think we can extract one method to remove the duplicated code.

components-jira/src/test/java/org/talend/components/jira/runtime/writer/JiraDeleteWriterTestIT.java
```diff
@@ -57,36 +74,49 @@ public void testWriteUnauthorized() throws IOException {
...
      */
-    @Test(expected = IOException.class)
+    @Test
     public void testWriteNotOpened() throws IOException {
-        JiraWriter deleteProjectWriter = writeOperation.createWriter(null);
+        IndexedRecord wrongProjectRecord = new GenericData.Record(DELETE_SCHEMA);
+        wrongProjectRecord.put(0, "WP");
+
+        thrown.expect(IOException.class);
+        thrown.expectMessage("Writer wasn't opened");
```
wwang-talend: i18n the exception message for some case, then assert the message, seems easy to break the test.

components-jira/src/main/java/org/talend/components/jira/connection/Rest.java
gonchar-ivan: in UNAUTHORIZED case Jira server return rubbish in http entity body. I don't want to provide this rubbish to user.
That's the reason of if-else block.

As you can see this if-else block is present for post, put, delete methods.
These methods should not return huge data. These methods should return only short message about status (which I provide to user in case of Exception)

As for get method:
Seems, I need to fix here, because get method supposed to return huge data.
In case of errors during get method I provide exceptions from apache http client library to user.
It's the 2nd reason why I don't need entityBody in get method.
I'll check it and fix. Thanks

components-jira/src/test/java/org/talend/components/jira/runtime/writer/JiraDeleteWriterTestIT.java
gonchar-ivan: Do you mean the case if message will come in some other language than English?

If yes, you are right. I'll try to add JUNIT assumption to check locale.
So, test will run only if English locale is chosen

components-jira/src/main/resources/org/talend/components/jira/runtime/writer/JiraWriter.properties
```diff
@@ -0,0 +1,17 @@
+error.writerNotOpened="Writer wasn''t opened"
```
wwang-talend: why double ''?

gonchar-ivan: It is for escaping. The same as \" for double quotes character in Java strings.

components-jira/src/main/resources/org/talend/components/jira/runtime/writer/JiraWriter.properties
```diff
@@ -0,0 +1,17 @@
...
+error.schemaNotContainId="Record schema doesn''t contain id field"
+error.schemaNotContainJson="Record schema doesn't contain json field"
```
wwang-talend: strange to me, one use double ', one use single quote

gonchar-ivan: It's my mistake. Need to fix it. Thanks

PR #342, opened 2016-11-29
fix(TCOMP-372): implements the JDBC changes

wwang-talend:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows Talend standard
- [x] Tests for the changes have been added (for bug fixes / features)

**What is the current behavior?** (You can also link to an open issue here)
https://jira.talendforge.org/browse/TCOMP-372

**What is the new behavior?**
https://jira.talendforge.org/browse/TCOMP-372

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

**Other information**:
https://jira.talendforge.org/browse/TCOMP-372

PR #345, opened 2016-11-29
Revert "fix(TCOMP-372): implements the JDBC changes"

gonchar-ivan:
Reverts Talend/components#342


gonchar-ivan:
Reverts https://github.com/Talend/components/pull/342

PR #363, opened 2016-12-08
feat(TCOMP-275): Implement multi-execution engine support.

RyanSkraba:

All definitions that might have a runtime instance to execute business logic implement the `RuntimableDefinition`.  There is no mechanism to specify whether the component (and runtime) support a particular execution engine.

**What is the new behavior?**

An enum was added `ExecutionEngine` that provides a list of execution engines that can be used to run a component.  The starting values are `DI` and `BEAM` (with `DI_SPARK_BATCH` and `DI_SPARK_STREAMING` reserved for future studio integration).

`ComponentDefinition`s *must* now specify which execution engines are possible and *may* provide a custom runtime for each supported engine.  

`DatastoreDefinition` and `DatasetDefinition` have always needed to run "standalone" (i.e. outside of an execution engine) and do not need  this mechanism.

The `RuntimableDefinition` was removed, since different types of definitions no longer have common requirements for providing the runtime.

- [X] Yes
- [ ] No

Since the `ComponentDefinition` *must* specify its valid execution engines, it is now mandatory in `AbstractComponentDefinition`.  A helper method was provided to throw an exception if an unsupported execution engine is requested.

**Other information**:

I assigned `ExecutionEngine.DI` **OR** `ExecutionEngine.BEAM` to components depending on their current use.  Some components that provide a TCOMP `Source` or `Sink` should be compatible with DI **AND** BEAM.  For the moment, I'm leaving this as future validation work.

If this PR is accepted, a Studio change will be necessary to specify the `ExecutionEngine` currently being used to run the component.


build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/110/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/110/)

components-api/src/main/java/org/talend/components/api/component/runtime/ExecutionEngine.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * When creating a {@link org.talend.daikon.runtime.RuntimeInfo} object via a
+ * {@link org.talend.components.api.component.ComponentDefinition}, you might want to specify where the runtime instance
+ * will be running.
+ */
+public enum ExecutionEngine {
+    DI,
+    DI_SPARK_BATCH,
+    DI_SPARK_STREAMING,
+    BEAM
```
bchen-talend: ExecutionEngine is equivalent to APP? will we use this value to know which component can be listed in which APP? if so do you think we need to separate BEAM to DS and DP? 
for example JDBCOutputDefinition do not supported by DP but DS.

Or we will have another catalog for APP, between APP and ExecutionEngine(or Implement Type), there is a mapping, such as for DS, it support two kind of implement, Beam implement or TCOMP Source/Sink implement, and return Beam implement first if exist, then TCOMP Source/Sink implement.
Define ComponentDefinition by APP name, getRuntimeInfo with ExecutionEngine Type(or Implement Type), the mapping controlled by global, such as ComponentService

build-talend:

components-api/src/main/java/org/talend/components/api/component/runtime/ExecutionEngine.java
francisu: @bchen-talend I'm not sure of the APP concept (that is, I don't know if I support doing something like that). But the execution engine is a very specific thing, it's the environment in which the component executes, which is one of the 4 environments. If we want to have the APP concept, it's a different matter. For connector components, there is no architectural difference between any of the execution environments. That is, any connector component that implements TCOMP source/sink shall work with the same exact runtime in any supported execution environment. 
As a separate matter, sometimes in Beam there might be a native Beam IO that the TCOMP component wants to use, so the TCOMP component can have a Beam-specific runtime that calls that native Beam component, but this is not required, and is a decision that can be made for each component. Strictly speaking, it's just an optimization. _The standard TCOMP source/sink runtime should work functionally exactly as well in the Beam environment._

components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
```diff
@@ -13,33 +13,38 @@
 package org.talend.components.api.component;
 
 import java.lang.reflect.Array;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
```
francisu: We don't do this "*" thing as far as I know.

components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
```diff
@@ -13,33 +13,38 @@
...
      * Constructor sets component name
      * 
      * @param componentName component name
+     * @param engine1 an execution engine that this component is valid for.
+     * @param engineOthers other execution engines that this component can generate runtimes for. (This technique with
+     * varargs is just to guarantee that there is at least one execution engine specified.)
      */
-    public AbstractComponentDefinition(String componentName) {
+    public AbstractComponentDefinition(String componentName, ExecutionEngine engine1, ExecutionEngine... engineOthers) {
         this.componentName = componentName;
```
francisu: I think by default all execution engines should be supported (this is the whole idea of TCOMP), and the only reason to specify the execution engines separately would be if you wanted to restrict it to a subset. Sorry, it looks like you did a lot of work to make it always explicit, but I think the change can be much simpler (and less disruptive). I would live this constructor as it is, and add a new one that supports the varargs execution engine.

components-api/src/main/java/org/talend/components/api/component/runtime/ExecutionEngine.java
francisu: @RyanSkraba Why does it turn out to be a good idea to force the component to indicate at lease one ExecutionEngine? (I read your comments carefully but could not see that point).

bchen-talend: thanks @francisu, the reason I added APP concept is to list supported components for different Application, but no need then, because I didn't know we will filter components by application exclusive TCOMP Service

components-api/src/main/java/org/talend/components/api/service/ComponentService.java
```diff
@@ -164,11 +165,12 @@
      * list all the depencencies required for this component to be executed at runtime
      * 
      * @param componentName name of the component to get the dependencies of.
+     * @param engine the execution engine that the component should run in.
      * @param properties the properties to compute the dependencies from.
      * @param componentType will determine the runtime class to be used.
      * @return the runtime info for running the given component with the given properties
      */
-    RuntimeInfo getRuntimeInfo(String componentName, Properties properties, ConnectorTopology componentType);
+    RuntimeInfo getRuntimeInfo(String componentName, ExecutionEngine engine, Properties properties, ConnectorTopology componentType);
```
sgandon: this will require some changes in the Studio as well.

sgandon: Don't you think we need some API to get the components for a given ExecutionEngine ? like the Studio could use this to filter which component to present on the palette ?(today they just embed the components they want manually). Even for datastream that would be a way to filter components. 
what do you think ?

components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
RyanSkraba: In any case, we have to touch all of the components anyway for the new signature of `ComponentDefinition#getRuntimeInfo()` which is a breaking change, so we're not saving much here by having a constructor that is implicitly all environments.

I DO agree that we should have a constructor that lets you specify ALL execution environments, and I will make this change!  Not for backwards compatibility, but for the TCOMP promise :)  I don't want to move a component from DI to ALL until there's someone responsible for validating it in an ExecutionEngine, though!

(The choice to limit the existing components to `ExecutionEngine.DI` in this PR is to limit the impact of validating all of the components in a Beam environment.)

components-api/src/main/java/org/talend/components/api/service/ComponentService.java
RyanSkraba: I know :( ... both for this Service and the changed `getRuntimeInfo()` methods.  I think it's a better idea to break backwards compatibility like this than to graft in a `DI`-only default.

components-api/src/test/java/org/talend/components/api/component/runtime/SimpleRuntimeInfoTest.java
```diff
@@ -67,29 +67,7 @@ public URLStreamHandler createURLStreamHandler(String protocol) {
...
-                return null;
-            }
-
-            @Override
-            public Set<ConnectorTopology> getSupportedConnectorTopologies() {
-                return null;
-            }
-        };
-        SimpleRuntimeInfo runtimeInfo = new SimpleRuntimeInfo(cd.getClass().getClassLoader(),
+        SimpleRuntimeInfo runtimeInfo = new SimpleRuntimeInfo(getClass().getClassLoader(),
```
sgandon: nice catch.

components-api/src/main/java/org/talend/components/api/component/runtime/ExecutionEngine.java
francisu: @RyanSkraba also the tSalesforceConnection is a very interesting example. Of course I agree it does not exist in the Beam environment. We need to provide some clean and simple way for the tSalesforceInput component to configure itself in both the DI and Beam environments where the connection mechanism is very different. Some more thought is necessary for this.

francisu: I fully agree with @sgandon about the APP concept.

components-api/src/main/java/org/talend/components/api/component/AbstractComponentDefinition.java
francisu: @RyanSkraba completely agree with you, and limiting the existing components to DI at this time makes sense (except for those components we expect to work on Beam). GIven the connection issues, it's clear the Salesforce components might have some issues on Beam at this time (they will probably work, if you do a tSalesforceInput and chose to embed the connection information in the component -- but that's not the TFD way and we want to do someting about that).

francisu: I do think that when we get further along with TFD and DI and TCOMP as far as certifying the execution environments and making sure we have the right abstractions for things like connection specifications, newer components will use the default of _all_ execution environments, as it will just work. But there is a verification/qa step that's required and should be done as a matter of course. That is, we don't wait for the verification step before specifying the execution environment. But we are not there yet, since there are issues with the execution environments and we are early.

RyanSkraba:
Hello!  I've made the change so that AbstractComponentDefinition has a constructor that accepts ALL ExecutionEngines, which _should_ be possible for all components with a runtime based on TCOMP Source or Sink.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/130/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/130/)

RyanSkraba:
Last build was aborted to free up executors.  Trying a rebuild now, then I'll merge.

PR #366, opened 2016-12-12
Feat/tdp 2048 better jdbc error handling

hmdebenque:

When user supply a wrong SQL query, the API returns an "unexpected exception"

**What is the new behavior?**
When user supply a wrong SQL query, the API returns a more detailed message from JDBC

- [ ] Yes
- [X] No

**Other information**:
There might be other changes needed in JDBC component where SQL syntax errors are not caught separately.


components-jdbc/src/main/java/org/talend/components/jdbc/runtime/JDBCSourceOrSink.java
```diff
@@ -89,6 +86,7 @@ private static ValidationResult fillValidationResult(ValidationResult vr, Except
     @Override
     public ValidationResult validate(RuntimeContainer runtime) {
         ValidationResult vr = new ValidationResult();
+        // TODO C'est moi ou la connection devrait être fermée là?
```
sgandon: yes absolutely can you fix it please ?

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/JDBCSourceOrSink.java
```diff
@@ -128,6 +126,8 @@ public Schema getSchemaFromQuery(RuntimeContainer runtime, String query) {
                 ResultSet resultset = statement.executeQuery(query)) {
             ResultSetMetaData metadata = resultset.getMetaData();
             return JDBCAvroRegistry.get().inferSchema(metadata);
+        } catch (SQLSyntaxErrorException sqlSyntaxException) {
+            throw new ComponentException(JdbcComponentErrorsCode.SQL_SYNTAX_ERROR, sqlSyntaxException);
         } catch (Exception e) {
```
sgandon: I hate those catch *Exception* could you check if it is required ?

services/components-api-service-rest/src/main/java/org/talend/components/service/rest/impl/ControllersConfiguration.java
```diff
@@ -39,30 +39,42 @@
...
 @ControllerAdvice
 public class ControllersConfiguration {
 
     private static final Logger log = LoggerFactory.getLogger(ControllersConfiguration.class);
 
     @ExceptionHandler
+    public ResponseEntity<ApiError> handleTalendExceptions(TalendRuntimeException e) {
+        log.info("A Talend exception reached the API", e);
+        Throwable cause = e.getCause();
+        return buildUnexpectedErrorResponse(e.getCode(), cause == null ? e.getMessage() : cause.getMessage(), cause);
```
sgandon: what is the point to have TalendRuntimeException if you strip it naked ? you may have parameters as well and other things.

hmdebenque: As the i18n must be done in TComp, the client only need the error code, the message and a title. However, the log message contains all relevant data about the exception.
The log level is info because they are many talend exception thrown as part of components API as the SQLSyntaxError wrapped in Component exception: it is expectable a user may type a wrong SQL statement and we should not spam error or warn level because of it.

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/JDBCSourceOrSink.java
hmdebenque: Yes this will only catch API advirtised exceptions: ClassNotFoundException when driver's not present and SQLException.

components-jdbc/src/main/java/org/talend/components/jdbc/runtime/JDBCSourceOrSink.java
hmdebenque: It seems that the connection, once open is stored in the runtime context and used in other places. Closing the connection here would break other parts.
It would take a refactoring of the JDBC component and maybe make use of connection pools.

build-talend:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/223/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/223/)

PR #375, opened 2016-12-15
fix(TFD-148): make jdbc definition return right runtime

bchen-talend:

JdbcInput/Output do not return Beam runtime

**What is the new behavior?**
JdbcInput/Output return Beam runtime if the execution engine is Beam

- [ ] Yes
- [x] No

**Other information**:


bchen-talend:
The error same as what Sébastien said
> there  is a issue with some IT tests because when they are launched the first time they expect the artifact they are testing to be installed but in maven the Integration Tests are launch before the install phase.

PR #376, opened 2016-12-15
fix(TDI-37973): adjust the pom to make jdbc components can work in studio

wwang-talend:

**What is the new behavior?**

- [ ] Yes
- [x] No

**Other information**:
https://jira.talendforge.org/browse/TDI-37973


wwang-talend:
after the fix, the Bundle-SymbolicName will be different for jdbc definition and runtime. But it's not enough for making jdbc work again in studio as they depend on components-adapter-beam lib which is not a plugin lib.

RyanSkraba:
Hello!  This looks good to me, noting the problem with the Beam API.  If we move the beam stuff to a separate module, this looks like it's enough to fix this, right?

PR #387, opened 2016-12-22
feat(TFD-142): refactor kafka schema

bchen-talend:

column key is mandatory

**What is the new behavior?**
remove raw format type, but add csv format.
for input, only return value of kafka record, ignore the key of kafka record.
for output, add partition key, defualt is round-robin, means send null for key of kafka record.
else user can choice one column of record as the key, but this column also a part of value for kafka record.

and for avro format, output component work with empty schema, which means use incoming data's schema

UI changes:
KafkaDataset:
it will keep `topic` and one empty schema(no fixed columns `key`, `value` anymore)
and add:
`value format` with `csv` & `avro` (support more later)
   when select csv
       set field delimiter
   when select avro
       option 1: avro schema generated by component's schema
       option 2: avro schema string provided by user

KafkaOutput:
add partition type
   1. round-robin
   2. one column from the record

- [x] Yes
- [ ] No

It break current KafkaDataset/Input/Output

**Other information**:


build-talend:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/381/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/381/)

components/components-kafka/kafka-definition/src/main/java/org/talend/components/kafka/dataset/KafkaDatasetProperties.java
```diff
@@ -131,27 +126,8 @@ public void setDatastoreProperties(KafkaDatastoreProperties datastoreProperties)
...
-                }
-                valueSchema = fields.endRecord();
-            }
-        }
-        return valueSchema;
-    }
-
     public enum ValueFormat {
-        RAW,
+        CSV,
```
bchen-talend: For the moment, I don't know how RAW format work with other components, the schema will be fixed and allow only one column.

components/components-kafka/kafka-runtime/src/main/java/org/talend/components/kafka/runtime/KafkaInputPTransformRuntime.java
```diff
@@ -12,28 +12,44 @@
...
 import org.talend.components.api.component.runtime.RuntimableRuntime;
 import org.talend.components.api.container.RuntimeContainer;
-import org.talend.components.kafka.dataset.KafkaDatasetProperties;
 import org.talend.components.kafka.input.KafkaInputProperties;
 import org.talend.daikon.properties.ValidationResult;
 
 public class KafkaInputPTransformRuntime extends PTransform<PBegin, PCollection<IndexedRecord>>
         implements RuntimableRuntime<KafkaInputProperties> {
 
+    static {
```
RyanSkraba: In `components-fileio` we discovered that this will not be called correctly when run on a cluster!  Deserializing a java static inner class does not call the static initializer on the parent class :P

As a quick workaround, can you move this initializer into one of the static inner classes that uses it?  ExtractCSVSplit is probably a good candidate!

components/components-kafka/kafka-definition/src/main/java/org/talend/components/kafka/dataset/KafkaDatasetProperties.java
```diff
@@ -41,25 +41,23 @@
...
-
-    // Fixed column name for value of kafka record
-    public final static String COLUMN_VALUE = "value";
-
     public ReferenceProperties<KafkaDatastoreProperties> datastore = new ReferenceProperties<>("datastore",
             KafkaDatastoreDefinition.NAME);
 
     public Property<String> topic = PropertyFactory.newString("topic");
 
+    public SchemaProperties main = new SchemaProperties("main");
```
RyanSkraba: I don't really understand the difference between `main` and `avroSchema`!  `main` is used unless `isHierarchy` is checked, but how do we expect it to be set differently than `avroSchema`?

Is this just the difference between using the schema editor (which doesn't exist in Data Streams yet) and manually specifying the schema as a JSON string?

components/components-kafka/kafka-runtime/src/main/java/org/talend/components/kafka/runtime/KafkaInputPTransformRuntime.java
```diff
@@ -72,7 +103,56 @@ public ValidationResult initialize(RuntimeContainer container, KafkaInputPropert
 
     @Override
     public Coder getDefaultOutputCoder() {
-        return KafkaIndexedRecordCoder.of();
+        return LazyAvroCoder.of();
```
RyanSkraba: If the schema is *known*, this should be just an `AvroCoder`!  The LazyAvroCoder should only be used if the schema is unknown until runtime.

components/components-kafka/kafka-runtime/src/main/java/org/talend/components/kafka/runtime/KafkaOutputPTransformRuntime.java
```diff
@@ -49,52 +45,37 @@
...
-                .withValueCoder(ByteArrayCoder.of()).updateProducerProperties(KafkaConnection.createOutputMaps(properties)));
+                .withTopic(properties.getDatasetProperties().topic.getValue())
+                .updateProducerProperties(KafkaConnection.createOutputMaps(properties));
 
+        switch (properties.partitionType.getValue()) {
+        case COLUMN: {
+            PCollection pc1 = objectPCollection.apply(WithKeys.of(new ProduceKey(properties.keyColumn.getValue())));
+            if (useAvro) {
+                // TODO for now use incoming avro schema directly, do not check configured schema, improvement it.
+                return (PDone) pc1.apply(kafkaWrite.withKeyCoder(ByteArrayCoder.of()).withValueCoder(LazyAvroCoder.of()));
```
RyanSkraba: Hmmm... looking at this, it might be really useful to use LazyAvroCoder to auto-detect the schema on write!  You can probably leave these.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/499/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/499/)

components/components-kafka/kafka-definition/src/main/java/org/talend/components/kafka/dataset/KafkaDatasetProperties.java
bchen-talend: another usage is for kafka output. when use column as key partition, need to select which column of `main` schema as the key.(yes, it's only work for csv and simple avro, not Hierarchy avro).
so how to let the user select one column from the hierarchy schema, suppose we will have a hierarchy schema wizard replace both `main` schema editor and `avroSchema` in future

PR #399, opened 2017-01-02
fix(fileio): static initialization of Avro converters

RyanSkraba:
Ensure that the static initialization for the fileio AvroRegistry occurs in an class that is serialized with the Beam job graph.

The static initialization for this class occurs in the runtime object -- unfortunately this is not the object that is serialized and accessed on the node, so the registered Avro converters are never available on the node.

**What is the new behavior?**

The static initializer is only added for the CSV file source, which is serialized across the network and is the only Beam object that requires it.

- [ ] Yes
- [X] No

**Other information**:

No unit tests were added -- there isn't any way to test this outside of an integration test.


pbailly:
LGTM

PR #402, opened 2017-01-03
fix: always activate snappy compression for dataprep avro.

RyanSkraba:

Generated avro files are not compressed.

**What is the new behavior?**

Generated avro files are compressed using the snappy codec

- [ ] Yes
- [X] No

**Other information**:


components/components-fileio/simplefileio-runtime/src/main/java/org/apache/beam/sdk/io/hdfs/ConfigureOnWriteHdfsFileSink.java
```diff
@@ -253,11 +263,14 @@ public void write(KV<K, V> value) throws Exception {
                 if (formatClass == (Class<?>) AvroKeyOutputFormat.class) {
                     AvroKey<IndexedRecord> k = (AvroKey<IndexedRecord>) value.getKey();
                     AvroJob.setOutputKeySchema(job, k.datum().getSchema());
+                    FileOutputFormat.setCompressOutput(job, true);
+                    job.getConfiguration().set(AvroJob.CONF_OUTPUT_CODEC, DataFileConstants.SNAPPY_CODEC);
                 }
```
pbailly: Please use an elseif here.
Except that, LGTM.

PR #404, opened 2017-01-04
fix(TDP-2739): Fix CSV on FileInput and Parquet on FileOutput

pbailly:
Allow CSV mode of FileInput to support empty column.
Correctly display the Byte Buffer on FileOutput.

File Input does not support CSV with empty field (or invalid number of field)
File Output does not display correctly the byte buffer

**What is the new behavior?**
Both problems are fixed.

- [ ] Yes
- [ ] No

**Other information**:
Warning, a bug still exist for the byteBuffer serialization/deserialization on beam direct runner. It will be automatically fixed as soon as beam is patched.


components/components-fileio/simplefileio-runtime/src/main/java/org/talend/components/simplefileio/runtime/SimpleFileIoOutputRuntime.java
```diff
@@ -185,7 +187,11 @@ public void processElement(ProcessContext c) {
             for (int i = 0; i < size; i++) {
                 if (sb.length() != 0)
                     sb.append(fieldDelimiter);
-                sb.append(in.get(i));
+                if (Schema.Type.BYTES.equals(in.getSchema().getFields().get(i).schema().getType())) {
```
RyanSkraba: Should use unwrapNullable for the field schema.

PR #447, opened 2017-01-26
feat(TDF-477): Utilities for generating random Avro rows.

RyanSkraba:

**What is the new behavior?**

The RowGeneratorIO can be used to generate random Avro records.

- [ ] Yes
- [X] No

**Other information**:


core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/GeneratorFunctions.java
```diff
@@ -0,0 +1,335 @@
...
+            }
+            byte[] buffer = new byte[size];
+            input.getRandom().nextBytes(buffer);
+            return new GenericData.Fixed(schema, buffer);
+        }
+    }
+
+    public static class BytesGenerator extends GeneratorFunction<ByteBuffer> {
+
+        private final int maxSize;
```
pbailly: maxSize is useless.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/GeneratorFunctions.java
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    public static class StringGenerator extends GeneratorFunction<String> {
+
+        public static final char[] ALPHANUMERIC = ("ABCDEFGHIJKLMNOPQRSTUVWXYZ" + "abcdefghijklmnopqrstuvwxyz" + "123456780")
+                .toCharArray();
+
+        private final char[] characters;
+
+        private final int size;
```
pbailly: size is useless

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
```diff
@@ -0,0 +1,348 @@
...
+        private BoundedRowGeneratorReader(BoundedRowGeneratorSource source) {
+            this.source = source;
+            this.ctx = GeneratorFunction.GeneratorContext.of(source.partitionId);
+        }
+
+        @Override
+        public boolean start() {
+            // Set the random seed once if it hasn't already been set.
+            if (source.spec.getSeed() == null) {
+                ctx.setRandom(new Random(System.currentTimeMillis() + source.partitionId));
```
pbailly: Could you add the possibility to define a seed?
It will help us a lot to create fixed tests.

Note that the seed should be different in any executor, so that may not be an easy task.

RyanSkraba: Hello!  You can set the exact seed with `RowGeneratorIO.read().withSeed(...)` to use a predefined seed. 

There might be a problem here if System.currentTimeMillis() *happens* to be the exactly the same on two executors -- and they'll generate duplicates.  Maybe I should multiply the partition ID by a large value?

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/775/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/775/)

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/GeneratorFunctions.java
RyanSkraba: I made it configurable in the constructor -- try to keep Codacy happy :)

Eventually, this should be configurable via properties in the specified schema.

core/components-adapter-beam/src/test/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIOTest.java
```diff
@@ -0,0 +1,101 @@
...
+    @Test
+    @Category(RunnableOnService.class)
+    public void testBasicDeterministic() throws Exception {
+        // Non-deterministic read, just do a count.
+        PCollection<IndexedRecord> output = pipeline.apply(RowGeneratorIO.read()
+                .withSchema(SampleSchemas.recordCompositesRequired()).withSeed(0L).withRows(95L).withPartitions(13));
+        PAssert.thatSingleton(output.apply("Count", Count.<IndexedRecord> globally())).isEqualTo(95L);
+        pipeline.run();
+
+        // TODO: we could test the generated records here.
```
imejia: Indeed, beam has function to assure that an object is contained in a PCollection (PAssert.containsInAnyOrder), that plus a seed would make it. And this is important to test to be sure that we have reproducible row generation (via the seed).

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/GeneratorFunction.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Random;
+
+import com.google.common.base.Function;
+
+/**
+ * The generator function is Serializable, and turns the current state of the GeneratorContext into an Avro datum.
+ *
+ * @param <T> The expected type of the datum.
+ */
+public abstract class GeneratorFunction<T> implements Serializable, Function<GeneratorFunction.GeneratorContext, T> {
```
RyanSkraba: Nice!  Thanks -- I'll go with `SerializableFunction` until we have a use that needs type info or display data.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/GeneratorFunctions.java
```diff
@@ -0,0 +1,335 @@
...
+
+        private final GeneratorFunction<?>[] inner;
+
+        private final String jsonSchema;
+
+        private transient Schema schema;
+
+        public RecordGenerator(Schema recordSchema) {
+            jsonSchema = recordSchema.toString();
+            this.schema = schema;
```
RyanSkraba: Fixed.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/GeneratorFunctions.java
```diff
@@ -0,0 +1,335 @@
...
+        private final int size;
+
+        private final String jsonSchema;
+
+        private transient Schema schema;
+
+        public FixedBytesGenerator(Schema fixedSchema) {
+            size = fixedSchema.getFixedSize();
+            jsonSchema = fixedSchema.toString();
+            this.schema = schema;
```
RyanSkraba: Fixed.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
```diff
@@ -0,0 +1,349 @@
...
+
+        /**
+         * Provide the schema for the generated rows.
+         *
+         * @param schema the Schema to use in the generated rows.
+         * @return the {@link Read} with the specified schema
+         */
+        public Read withSchema(Schema schema) {
+            this.schema = schema;
+            checkArgument(schema != null, "RowGeneratorIO.read().withSchema(schema) called with null schema");
```
RyanSkraba: Fixed.

core/components-adapter-beam/src/test/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIOTest.java
RyanSkraba: I'm still thinking about a way to represent "expected rows" nicely for tests -- this will be useful for unit tests for other components as well.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/794/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/794/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/798/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/798/)

sgandon:
I really would have liked this row generator to be used for test any kind of runtime and not only beam once. Could this be also used for any kind of components ?

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
```diff
@@ -0,0 +1,349 @@
...
+            // The desiredBundleSizeBytes is ignored to split the rows into the bundles specified by the spec.
+            if (partitionId != -1) {
+                // If it has already been split, it can't be split further.
+                return Arrays.asList(this);
+            }
+
+            List<BoundedRowGeneratorSource> bundles = new ArrayList<>(spec.getPartitions());
+
+            // Evenly divide the rows over all the bundles but the last.
+            double rowsPerBundle = (double) spec.getRows() / spec.getPartitions();
```
jbonofre: For the Spark runner, the desiredBundleSizeBytes is computed using the estimatedSizeBytes and number of executor.
Why not using desiredBundleSizeBytes in combination with number of partition provided by user ?
Generally speaking, I would rather use only the desiredBundleSizeBytes by default instead of an user value.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
```diff
@@ -0,0 +1,349 @@
...
+            long startRowId = 0;
+            for (int partition = 0; partition < spec.getPartitions() - 1; partition++) {
+                long newRows = (long) (rowsPerBundle * (partition + 1)) - startRowId;
+                bundles.add(new BoundedRowGeneratorSource(spec, partition, startRowId, newRows));
+                startRowId += newRows;
+            }
+
+            // Put all of the remaining rows in the last bundle.
+            bundles.add(new BoundedRowGeneratorSource(spec, spec.getPartitions() - 1, startRowId, spec.getRows() - startRowId));
```
jbonofre: SourceUtils provides key range util to split. Not required here, just could be interesting.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
```diff
@@ -0,0 +1,349 @@
...
+            return bundles;
+        }
+
+        @Override
+        public long getEstimatedSizeBytes(PipelineOptions options) throws IOException {
+            return -1;
```
jbonofre: No way to estimate the size ? Using the number of row and a sampling of the row size (in bytes) for instance ?

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        @Override
+        public PCollection<IndexedRecord> expand(PBegin input) {
+            return input.apply(org.apache.beam.sdk.io.Read.from(new BoundedRowGeneratorSource(this)));
+        }
+
+        @Override
+        public void validate(PBegin input) {
+            checkState(getJsonSchema() != null, "RowGeneratorIO.read() requires a schema" + " to be set via withSchema(schema)");
```
jbonofre: rows and partitions seem mandatory too and you should check the state (even if the default read() defines it by default).

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorSchemaConstants.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Annotations that can be applied to Avro {@link org.apache.avro.Schema}s to help control the
+ */
+public abstract class RowGeneratorSchemaConstants {
+
+    // TODO(rskraba): Control the generation of fields using Avro annotations. None of this is implemented.
+
+    public final static String ROWGEN_NUMERIC_ROW_ID = "ROWGEN_NUMERIC_ROW_ID";
+
+    public final static String ROWGEN_NUMERIC_PARTITION_ID = "ROWGEN_NUMERIC_PARTITION_ID";
```
jbonofre: Could be directly in the IO  class right ?

core/components-adapter-beam/src/test/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIOTest.java
```diff
@@ -0,0 +1,101 @@
...
+        assertThat("All rows generated", expectedRowId, is(95L));
+    }
+
+    @Test
+    @Category(RunnableOnService.class)
+    public void testBasicDeterministic() throws Exception {
+        // Non-deterministic read, just do a count.
+        PCollection<IndexedRecord> output = pipeline.apply(RowGeneratorIO.read()
+                .withSchema(SampleSchemas.recordCompositesRequired()).withSeed(0L).withRows(95L).withPartitions(13));
+        PAssert.thatSingleton(output.apply("Count", Count.<IndexedRecord> globally())).isEqualTo(95L);
```
jbonofre: Counting is good, but a DoFn with assert of each element in the PCollection could be useful too.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
RyanSkraba: It doesn't seem right to return an estimated size here -- the doc says this is "in terms of external storage size, before any decompression or other processing done by the reader." and the `RowGeneratorIO` doesn't have an underlying storage.

In other words, it's used to generate the initial bundle splits, not meant to estimate the amount of outgoing data the bundle generates.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
RyanSkraba: I *would* agree if we were interested in optimizing the performance of the row generator `BoundedSource` -- that would be the technique to use.

In this case, however, I think that it's much more important that we can control and eventually "shape" the initial bundles manually through the read configuration.  The "skewed partition" behaviour is not yet implemented, but we would use it to check the behaviour of a downstream component.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
RyanSkraba: `SourceUtils` ?

core/components-adapter-beam/src/test/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIOTest.java
RyanSkraba: I'll put a TODO!  I want to stabilize the row generator a bit before testing for specific record data (it's supposed to be deterministic, as long as there aren't any changes to the generator functions, which might occur as we add features.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorSchemaConstants.java
RyanSkraba: This is a detail that needs to be worked out ... I'll delete this class from the PR for now.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
RyanSkraba: Done!

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/850/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/850/)

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
jbonofre: Understood. Thanks for the update.

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
jbonofre: Sorry, I mean that, related to my previous comment, you can use `ByteKeyRange` for the split. In your case, as you directly use the RowId (and not splitting by data volume), it's not useful.
The `SourceTestUtils` provides convenient ways to test your splitting.

Just for the record: https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/testing/SourceTestUtils.java

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
jbonofre: `RowGeneratorIO` could be considered as an external storage: it's basically the size of the data that the source will produce. It's used by the runners (for no-op runners like dataflow to bootstrap the corresponding number of machines) and in the split.
In your case, as your split doesn't use the estimated size, you don't care about the estimated size.

core/components-adapter-beam/src/test/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIOTest.java
jbonofre: It makes sense. Thanks !

core/components-adapter-beam/src/main/java/org/talend/components/adapter/beam/io/rowgenerator/RowGeneratorIO.java
RyanSkraba: For info -- we spoke about this and we'll make it possible to let Beam decide the number of partitions by setting the value to 0, which means that we need to sample the data to guess the record size.  This will be done in a future PR, however.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/858/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/858/)

PR #466, opened 2017-02-09
fix(TCOMP-476): Ensure that the singleton DefinitionRegistry is locked.

RyanSkraba:

The singleton DefinitionRegistry is not "locked" before sharing, so multithread use can cause failures.

**What is the new behavior?**

The getDefinitionRegistry() method is threadsafe.

- [ ] Yes
- [X] No

**Other information**:


services/components-api-service-spi/src/main/java/org/talend/components/service/spi/ServiceSpiFactory.java
```diff
@@ -13,32 +13,39 @@
...
-            defReg = new DefinitionRegistry();
-            for (ComponentInstaller installer : ServiceLoader.load(ComponentInstaller.class)) {
-                installer.install(defReg);
-            }
-        }
-        return defReg;
+  public static DefinitionRegistry getDefinitionRegistry() {
+    if (defReg == null) {
+      // Create a new instance of the definition registry.
+      DefinitionRegistry reg = new DefinitionRegistry();
```
sgandon: hummm, it looks like 2 (or more) registries can be created here before is field *defReg* is assigned

RyanSkraba: Not a problem since they're immutable once locked!  No reason to implement a complex double-checked locking here, which would be a bad idea if/when we ever want to implement reloadable registries.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/948/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/948/)

services/components-api-service-spi/src/main/java/org/talend/components/service/spi/ServiceSpiFactory.java
pradasouvanlasy: If we are certain that the _for_ loop below does not take too long to execute, then I guess that the scenario @sgandon mentioned should not have a significant impact on the performance. 
If you happend to have some concerns regarding this loop, I would prefer to do the `DefinitionRegistry` creation in a static block as already discussed with @RyanSkraba

RyanSkraba: Updated to initialize the definition registry singleton when the class is loaded.  This avoids any possible thread issues (until registry reloading is implemented).

services/components-api-service-spi/src/main/java/org/talend/components/service/spi/ServiceSpiFactory.java
```diff
@@ -24,10 +24,17 @@
...
 
+  /** Singleton for definition registry.  This will be reloaded if set to null. */
   private static DefinitionRegistry defReg;
 
+  /** Singleton for the component service.  This will be reloaded if set to null. */
   private static ComponentService componentService;
 
+  static {
+    // Ensure that the definition registry is created when the class is loaded.
+    getDefinitionRegistry();
```
sgandon: nice way to synchronize the creation.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/953/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/953/)

PR #481, opened 2017-02-16
fix(TCOMP-236): fix input component archetype

gonchar-ivan:

input-components-archetype is not correct

**What is the new behavior?**

archetype was cleaned up and fixed

- [ ] Yes
- [x] No

**Other information**:
https://jira.talendforge.org/browse/TCOMP-236

It's better to check generated project rather that this PR:
https://github.com/Talend/components/compare/work/TCOMP-236_input_archetype_example?expand=1

or in branch
https://github.com/Talend/components/tree/work/TCOMP-236_input_archetype_example/components/components-table

If you want try to build archetype do following:
mvn clean install (on components\examples\components-api-archetypes\input-component-archetypeinput-component)

than use following command to generate project:
mvn org.apache.maven.plugins:maven-archetype-plugin:2.4:generate -DarchetypeGroupId=org.talend.components -DarchetypeArtifactId=input-component-archetypes -DarchetypeVersion=0.18.0-SNAPSHOT


build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/192/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/192/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/283/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/283/)

gonchar-ivan:
Build failed not because of my changes

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/357/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/357/)

examples/components-api-archetypes/input-component-archetype/src/main/resources/archetype-resources/__rootArtifactId__-runtime/src/main/java/runtime/reader/__componentName__Reader.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @Override
+    public boolean advance() throws IOException {
+        if (!started) {
+            throw new IllegalStateException("Reader wasn't started");
+        }
+        hasMore = reader.ready();
+        if (hasMore) {
+            String line = reader.readLine();
+            IndexedRecord dataRecord = getConverter(line).convertToAvro(line);
```
sgandon: this is strange here.
I would have imagined you'd do something like (some pseudo code)
```
//create the data schema
Schema dataSchema = getRecordSchema(line); <- create the schema only once for all the records (some code removed here)
//create the data IndexRecord
IndexRecord dataRecord = getConverter(recordSchema).convertToAvro(line);
//create the outOfboand record (since the schema is static)
IndexedRecord outOfBandRecord = new GenericData.Record(${componentName}Properties.outOfBandSchema);
outOfBandRecord.put(0, result.totalCount);
then create the root IndexRecord
current = OutOfBandUtils.createRootRecord(dataRecord, outOfBandRecord);
```
To looks me this is more clear to me. And the `OutOfBandUtils.createRootRecord` would be in daikon too.
What do you think ?

examples/components-api-archetypes/input-component-archetype/src/main/resources/archetype-resources/__rootArtifactId__-runtime/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,7 @@
+log4j.rootLogger=DEBUG, stdout
```
sgandon: could you set it to INFO instead of DEBUG this is causing too much un-useful  logs in the build.

examples/components-api-archetypes/input-component-archetype/src/main/resources/archetype-resources/__rootArtifactId__-definition/src/main/java/__componentPackage__/__componentName__Properties.java
```diff
@@ -0,0 +1,226 @@
...
+     * 
+     * Main schema could be changes by user. So method should reconstruct Root schema each time
+     * 
+     * @param connector token to get the associated schema
+     * @param isOutputConnection whether the connection is an outgoing or incoming one
+     * @return Root schema, which contains main schema related to connector and static Out of band schema 
+     */
+    @Override
+    public Schema getSchema(Connector connector, boolean isOutputConnection) {
+        // design-time main schema associated with specified connector
```
gonchar-ivan: hm... I just realized _isOutputConnection_ parameter is not used here (and it is not required).
Schema fully defined by connector parameter (Input and Output schema have different connector instances associated with)
_isOutputConnection_ is used in getAvailableConnectors() to specify which kind of connector caller requested.
However, I'll add if statement for the moment

examples/components-api-archetypes/input-component-archetype/src/main/resources/archetype-resources/__rootArtifactId__-runtime/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+
+		<!-- APIs for all Components -->
+		<dependency>
+			<groupId>org.talend.components</groupId>
+			<artifactId>components-api</artifactId>
+			<version>${project.parent.version}</version>
+		</dependency>
+
+		<!-- Shared component properties such as Use/Password or Schema -->
+		<dependency>
```
gonchar-ivan: Agree about components-common. It is not really used in FileInput Runtime module.

And I'm not sure about components-api. My point of view is if some module directly uses some dependency (runtime module imports api classes), this dependency (even transitive) should be explicitly declared in module's pom.xml 
And runtime module directly uses components-api dependency.

Found following answers about the topic:
http://stackoverflow.com/questions/4226756/maven-should-i-keep-or-remove-declared-dependencies-that-are-also-transitives
http://stackoverflow.com/questions/4226756/maven-should-i-keep-or-remove-declared-dependencies-that-are-also-transitives

examples/components-api-archetypes/input-component-archetype/src/main/resources/archetype-resources/__rootArtifactId__-runtime/src/main/java/avro/StringConverter.java
```diff
@@ -0,0 +1,33 @@
+// ============================================================================
```
gonchar-ivan: 1) My first idea was that each component has a set of converters which are specific only for that component.
However, seems I was wrong. My statement is reasonable for top-level converter (top-level means datum object to IndexedRecord), but converters for each field may have similar logic. So, I agree to move them to daikon 

2) Just a note about package naming conventions in Daikon.
For converters we may have 2 approaches:
org/talend/daikon/avro/converter/{datumType} e.g org/talend/daikon/avro/converter/string
org/talend/daikon/avro/converters/{avroType} e.g. org/talend/daikon/avro/converter/logicalTime

first package name means this package it contains converters from datum string to (all) avro types
second package name means this package it contains converters from any datum type to avro logicalTime

I don't know which approach is best:
in datum approach we may have potentially infinite number of datum types -> thus packages
(however, surely we will not put all converters to daikon, because I think some of them will be really component specific)

examples/components-api-archetypes/input-component-archetype/src/main/resources/archetype-resources/__rootArtifactId__-runtime/src/test/resources/log4j.properties
gonchar-ivan: done in both log4j.properties: definition and runtime (will be done for integration module too, if such module required)

examples/components-api-archetypes/input-component-archetype/src/main/resources/archetype-resources/__rootArtifactId__-runtime/src/main/java/runtime/reader/__componentName__Reader.java
gonchar-ivan: Several notes:
1) Agree, Root record creation is a little bit dirty. So, better to have 1 code line instead of following 3:
```
current = new GenericData.Record(getRootSchema(line));
current.put(0, dataRecord);
current.put(1, outOfBandRecord);
```
2) However, I don't want to recreate Root Schema each time. I want to be able to cache it.
OutOfBandUtils.createRootRecord(dataRecord, outOfBandRecord) call supposes Root Schema will be recreated for each new record (from dataRecord, outOfBandRecord schemas)
3) As I remember, we discussed that in future we may have something else besides data record and outOfBand record inside root record (3 and more IndexedRecords inside one Root record). So, as for me this Util class should have RootRecordUtils name.
4) Another idea is to extend IndexedRecord. We may implement _RootRecord extends IndexedRecord_.
Its constructor should accept Root Schema and all nested records (data and outOfBand for the moment)
5) Why daikon and not components-api? I thought IndexedRecords are used only in components.
(I realized I made a mistake when I put RootSchema/RootRecordUtils in component-common. So, I think I should move to components-api, but I'm not sure it is needed in daikon)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/75/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/75/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/16/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/16/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/17/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/17/)

examples/components-api-archetypes/input-component-archetype/src/main/resources/archetype-resources/__rootArtifactId__-definition/src/main/java/StringDelimiter.java
```diff
@@ -0,0 +1,47 @@
...
+package ${package};
+
+/**
+ * Enumeration of standard string delimiters
+ * 
+ * This shows possible values for dropdown list UI element i18n message for it
+ * should be placed in TableInputProperties.properties file
+ */
+public enum StringDelimiter {
+	SEMICOLON {
```
sgandon: You could make it more simple like
SEMINCOLON(";"), COLON(",")...
private delim;
public StringDelimiter(String delim){
    this.delim = delim;
}
public getDelimiter(){
    return delim;
}

gonchar-ivan: got it. Thanks. fixing

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/19/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/19/)

examples/components-api-archetypes/input-component-archetype/src/test/resources/projects/archetypeTest/goal.txt
```diff
@@ -1 +1 @@
-clean verify clean
+clean install clean
```
sgandon: why did you change from verify to install ? this is the maven goal launched when testing the archetype creation ?

examples/components-api-archetypes/input-component-archetype/src/main/resources/archetype-resources/__rootArtifactId__-definition/src/main/java/RuntimeInfoProvider.java
```diff
@@ -0,0 +1,46 @@
...
+package ${package};
+
+import org.talend.components.api.component.runtime.DependenciesReader;
+import org.talend.components.api.component.runtime.JarRuntimeInfo;
+import org.talend.daikon.runtime.RuntimeInfo;
+
+/**
+ * Service class, which provides {@link RuntimeInfo} instances
+ */
+public final class RuntimeInfoProvider {
```
sgandon: I was wondering if we really need a class of this, I would rather be using the FamillyDefinition class.

examples/components-api-archetypes/input-component-archetype/src/test/resources/projects/archetypeTest/goal.txt
gonchar-ivan: this goal is executed on project generated from archetype.
verify -> install for integration tests. Otherwise IT in OSGi container fails

examples/components-api-archetypes/input-component-archetype/src/main/resources/archetype-resources/__rootArtifactId__-definition/src/main/java/RuntimeInfoProvider.java
gonchar-ivan: Definition classes uses Properties classes
see getPropertyClass() method

When I add some design time feature and need to get RuntimeInfo in Properties class.
In case of using FamilyDefinition as RuntimeInfo provider, Properties class will use FamilyDefinition.

Thus, something like cycle dependency (however, inside package or module only).

I wanted to avoid this.
I saw both Definition and Properties classes require RuntimeInfo.
Thus, I introduce separate class, which provides it.
I'm not sure about this decision.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/85/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/85/)

PR #482, opened 2017-02-17
fix(BEAM-1205): Always use @Rule for TestPipeline.

RyanSkraba:

BEAM-1205 was merged last night and breaks current Beam unit tests.

**What is the new behavior?**

Tests are no longer broken.

- [ ] Yes
- [X] No

**Other information**:


components/components-elasticsearch/elasticsearch-runtime_2_4/src/test/java/org/talend/components/elasticsearch/runtime_2_4/ElasticsearchBeamRuntimeTestIT.java
```diff
@@ -54,6 +54,9 @@
 
     Client client;
 
+    @Rule
+    public TestPipeline pipeline = TestPipeline.create();
```
imejia: The new new new standard is to call pipelines in tests p :P

RyanSkraba:
I've validated the previously failing tests with a local build while repo.spring.io is down and our CI can't progress.  Master is broken without these fixes, so I'm merging early.

PR #546, opened 2017-03-03
bugfix(TDI-38286): fix DATE and TIME type value through Dynamic

TotalP:

DATE and TIME type will get the incorrect value through Dynamic

**What is the new behavior?**
DATE and TIME type will get correct value through Dynamic

- [ ] Yes
- [x] No

**Other information**:
https://jira.talendforge.org/browse/TDI-38286

### DIFF
--- a/components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeWriter.java
+++ b/components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeWriter.java
@@ -18,12 +18,12 @@ import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.text.SimpleDateFormat;
-import java.util.List;
 import java.util.ArrayList;
-import java.util.Map;
-import java.util.HashMap;
-import java.util.Date;
 import java.util.Calendar;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.TimeZone;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -42,12 +42,12 @@ import org.talend.daikon.avro.AvroUtils;
 import org.talend.daikon.avro.SchemaConstants;
 import org.talend.daikon.avro.converter.IndexedRecordConverter;
 
+import com.snowflake.client.loader.LoadResultListener;
+import com.snowflake.client.loader.LoaderFactory;
+import com.snowflake.client.loader.LoaderProperty;
 import com.snowflake.client.loader.LoadingError;
 import com.snowflake.client.loader.Operation;
 import com.snowflake.client.loader.StreamLoader;
-import com.snowflake.client.loader.LoadResultListener;
-import com.snowflake.client.loader.LoaderProperty;
-import com.snowflake.client.loader.LoaderFactory;
 
 public final class SnowflakeWriter implements WriterWithFeedback<Result, IndexedRecord, IndexedRecord> {
 
--- a/components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeWriter.java
+++ b/components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeWriter.java
@@ -18,7 +18,13 @@ import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.text.SimpleDateFormat;
-import java.util.*;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.HashMap;
+import java.util.Date;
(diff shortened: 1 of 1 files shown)


components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeWriter.java
```diff
@@ -18,13 +18,7 @@
...
 import java.sql.SQLException;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Calendar;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.TimeZone;
+import java.util.*;
```
gonchar-ivan: We have a convention do not use asterisk in imports.
Please, replace java.util.*; with explicit import declarations (as it was before)

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeWriter.java
```diff
@@ -42,12 +36,7 @@
 import org.talend.daikon.avro.SchemaConstants;
 import org.talend.daikon.avro.converter.IndexedRecordConverter;
 
-import com.snowflake.client.loader.LoadResultListener;
-import com.snowflake.client.loader.LoaderFactory;
-import com.snowflake.client.loader.LoaderProperty;
-import com.snowflake.client.loader.LoadingError;
-import com.snowflake.client.loader.Operation;
-import com.snowflake.client.loader.StreamLoader;
+import com.snowflake.client.loader.*;
```
gonchar-ivan: The same here

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeWriter.java
```diff
@@ -326,8 +315,18 @@ public void write(Object datum) throws IOException {
         }
         IndexedRecord input = factory.convertToAvro(datum);
         List<Schema.Field> fields = input.getSchema().getFields();
+
+        // input and mainSchema synchronization. Such situation is useful in case of Dynamic
```
gonchar-ivan: Write() method is called multiple times. (It may be called a lot of time)
Result of below code will be the same across write() method calls.
I suggest to add new boolean field isFirst or firstRecord (is it a first record processed ) and create collectedFields in if statement:

```
if(isFirst) {
 collectedFields = ... ;
 isFirst = false;
}
```

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeWriter.java
```diff
@@ -326,8 +315,18 @@ public void write(Object datum) throws IOException {
...
+        List<Schema.Field> collectedFields = new ArrayList<>();
+        for (Schema.Field item : fields) {
+            Schema.Field fieldFromMainSchema = mainSchema.getField(item.name());
```
gonchar-ivan: _mainField_ is shorter name with the same meaning

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeWriter.java
gonchar-ivan: I don't get how this synchronization works. We need to discuss it.

TotalP: "Synchronization" does not mean java synchronization mechanism working with threads. In this context it means that fields in both schemas are the same

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeWriter.java
```diff
@@ -42,12 +42,12 @@
 import org.talend.daikon.avro.SchemaConstants;
 import org.talend.daikon.avro.converter.IndexedRecordConverter;
 
-import com.snowflake.client.loader.LoadResultListener;
-import com.snowflake.client.loader.LoaderFactory;
-import com.snowflake.client.loader.LoaderProperty;
 import com.snowflake.client.loader.LoadingError;
 import com.snowflake.client.loader.Operation;
 import com.snowflake.client.loader.StreamLoader;
+import com.snowflake.client.loader.LoadResultListener;
```
gonchar-ivan: One more bug from idea. It reorders imports (((

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/278/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/278/)

PR #551, opened 2017-03-06
feat(webapp):use Main form as default when none specified

sgandon:

when no form name is specified in the rest end points null was used.

**What is the new behavior?**
now the Form.MAIN is being used as default.

- [ ] Yes
- [X] No


sgandon:
please wait as I have forgotten other APIs

PR #560, opened 2017-03-09
fix(TCOMP-498):fix and simplify code and unit test for kafka topic

sgandon:

the topic field was using NamedThing for possible value but did not use i18n, so there is not need to use NamedThing and string can be used instead

**What is the new behavior?**
using String for kafka topics instead of named thing.

- [ ] Yes
- [X] No


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/351/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/351/)

PR #575, opened 2017-03-15
bugfix(TDI-38272) : fix NPE when connection is closed

TotalP:

- [-] Tests for the changes have been added (for bug fixes / features)

NPE when connection is closed

**What is the new behavior?**
Error message is more informative

- [] Yes
- [x] No

**Other information**:
https://jira.talendforge.org/browse/TDI-38272

### DIFF
--- a/components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeSourceOrSink.java
+++ b/components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeSourceOrSink.java
@@ -15,6 +15,19 @@
  */
 package org.talend.components.snowflake.runtime;
 
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.Driver;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.DatabaseMetaData;
+import java.sql.DriverPropertyInfo;
+import java.sql.SQLFeatureNotSupportedException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+
 import org.apache.avro.Schema;
 import org.apache.avro.Schema.Field;
 import org.slf4j.Logger;
@@ -29,12 +42,6 @@ import org.talend.daikon.SimpleNamedThing;
 import org.talend.daikon.avro.SchemaConstants;
 import org.talend.daikon.properties.ValidationResult;
 import org.talend.daikon.properties.ValidationResult.Result;
-
-import java.io.IOException;
-import java.sql.*;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Properties;
 
 public class SnowflakeSourceOrSink implements SourceOrSink {
 
--- a/components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeSourceOrSink.java
+++ b/components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeSourceOrSink.java
@@ -15,19 +15,6 @@
  */
 package org.talend.components.snowflake.runtime;
 
-import java.io.IOException;
-import java.sql.Connection;
-import java.sql.SQLException;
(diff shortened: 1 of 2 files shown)


components/components-snowflake/src/test/java/org/talend/components/snowflake/runtime/SnowflakeSourceOrSinkTest.java
```diff
@@ -0,0 +1,58 @@
...
+public class SnowflakeSourceOrSinkTest {
+    @Mock
+    private RuntimeContainer runtimeContainerMock = Mockito.mock(RuntimeContainer.class);
+
+    private SnowflakeSourceOrSink snowflakeSourceOrSink;
+
+    @Before
+    public void setUp() throws Exception {
+        this.snowflakeSourceOrSink = new SnowflakeSourceOrSink();
+        this.snowflakeSourceOrSink.properties = new SnowflakeConnectionProperties("name") {
```
iistomin-gl: Why properties is set via direct field access ? You can call SnowflakeSourceOrSink.initialize(RuntimeContainer, ComponentProperties) method.

iistomin-gl: Instead of overriding of getReferenceComponentId() method you could do something like this:
```java
SnowflakeConnectionProperties properties = new SnowflakeConnectionProperties("test");
properties.referencedComponent.componentInstanceId.setValue("referencedComponentId");
this.snowflakeSourceOrSink.initialize(runtimeContainerMock, properties);
```

TotalP: Ok, I've got it

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/33/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/33/)

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeSourceOrSink.java
```diff
@@ -43,6 +30,12 @@
 import org.talend.daikon.properties.ValidationResult;
 import org.talend.daikon.properties.ValidationResult.Result;
 
+import java.io.IOException;
+import java.sql.*;
```
iistomin-gl: As far as I remember, wildcard import (*) is not allowed by Talend code conventions/style.

components/components-snowflake/src/test/java/org/talend/components/snowflake/runtime/SnowflakeSourceOrSinkTest.java
```diff
@@ -0,0 +1,53 @@
...
+        properties.referencedComponent.componentInstanceId.setValue("referencedComponentId");
+        this.snowflakeSourceOrSink.initialize(runtimeContainerMock, properties);
+    }
+
+    /**
+     * Checks {@link SnowflakeSourceOrSink#connect(RuntimeContainer)} throws {@link IOException} when connection in null
+     */
+    @Test(expected = IOException.class)
+    public void testConnectWhenConnectionIsNull() throws Exception {
+        Mockito.when(runtimeContainerMock.getComponentData(Matchers.anyString(), Matchers.anyString())).thenReturn(null);
```
iistomin-gl: Just a recommendation, use expected values for matching where possible, not just _anyString_. I would wrote something like this:

```java
Mockito.when(runtimeContainerMock.getComponentData(Matchers.eq("referencedComponentId"), Matchers.eq(SnowflakeSourceOrSink.KEY_CONNECTION))).thenReturn(null);
```

TotalP: Ok, I think in this case it does not matter what String value for matching would be. Test covers null - connection only.

PR #576, opened 2017-03-15
feat(TDI-38190) : add fetchsize property

dmytro-sylaiev:
* Add 2 properties into SnowflakeOutputProperties: isFetchSizeUsed and
* fetchSize (second is hidden)

* If fetchSize is enabled - add it into Snowflake writer api

- [+] The commit message follows Talend standard
- [-] Tests for the changes have been added (for bug fixes / features)

**What is the new behavior?**

- [ ] Yes
- [+] No


components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakeoutput/TSnowflakeOutputProperties.java
```diff
@@ -110,14 +120,24 @@ public void setupLayout() {
...
+        advForm.getWidget(fetchSize).setHidden();
+        
     }
 
     public void afterOutputAction() {
         refreshLayout(getForm(Form.MAIN));
     }
+    
+    public void afterIsFetchSizeUsed() {
+		getForm(Form.ADVANCED).getWidget(fetchSize.getName()).setHidden(!isFetchSizeUsed.getValue());
```
sgandon: it is more that advised to change the layout in the refreshLayout method that can be called by the framework at different moment in the life of the properties. could you move this code to the `refreshLayout()` method please ?

PR #599, opened 2017-03-21
feat(TCOMP-508): add RootRecordUtils.createRootRecord(Schema rootSche…

gonchar-ivan:
…ma, IndexedRecord dataRecord, IndexedRecord outOfBandRecord) method

**What is the new behavior?**

added new method

- [ ] Yes
- [x] No

**Other information**:

https://jira.talendforge.org/browse/TCOMP-508


core/components-common/src/main/java/org/talend/components/common/component/runtime/RootRecordUtils.java
```diff
@@ -63,6 +63,23 @@ public static IndexedRecord createRootRecord(Schema rootSchema) {
...
+     * Creates and returns root record
+     * For performance purposes, it is recommended to create Root schema only once and then use it to create Root record instances
+     * Sets <code>dataRecord</code> on position 0, and <code>outOfBandRecord</code> on position 1
+     * 
+     * @param rootSchema Root schema, see {@link RootSchemaUtils}
+     * @param dataRecord main data record
+     * @param outOfBandRecord out of band (additional) data record
+     * @return root record
+     */
+    public static IndexedRecord createRootRecord(Schema rootSchema, IndexedRecord dataRecord, IndexedRecord outOfBandRecord) {
```
sgandon: I would expect here that the rootSchema is not used as a parameter but created on the fly in this method so that all Root Schemas follow the same pattern especially with field namings that should be constants.
something like
```
     Schema rootSchema = SchemaBuilder.record("Root").fields() //
                .name("Main").type(dataRecord.getSchema()).noDefault() // 
                .name("OutOfBand").type(outOfBandRecord.getSchema()).noDefault() // 
                .endRecord(); //
```

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/56/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/56/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/60/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/60/)

core/components-common/src/main/java/org/talend/components/common/component/runtime/RootRecordUtils.java
gonchar-ivan: hm... right. All root schemas should have the same pattern.

However, I didn't want to create new schema instance for each record. Is it reasonable? Or it's not a problem?
If it is not a problem, I will use RootSchemaUtils.createRootSchema() method here.
(Seems, I should not care about additional instance creation until I haven't check it with profiler. I may think that it is a problem, but it is really not )

 I thought user will use RootSchemaUtils.createRootSchema() method to create schema.
Then this root schema will be cached in component.
Component passes root schema as an argument each time it creates new root record.

I may add RootSchemaUtils.isRootSchema() check in createRootRecord()

sgandon: Ok I agree with this approach, this allow for the user to cache the root schema.

gonchar-ivan: I realized I have already added isRootSchema() checking )
So, if user passes not root schema method will throw IllegalArgumentException.

core/components-common/src/test/java/org/talend/components/common/component/runtime/RootRecordUtilsTest.java
```diff
@@ -139,6 +140,36 @@ public void testCreateRootRecordNotRootSchema() {
...
+        Schema rootSchema = SchemaBuilder.record("Root").fields() //
+                .name("Main").type(mainSchema).noDefault() // 
+                .name("OutOfBand").type(outOfBandSchema).noDefault() // 
+                .endRecord(); //
+        
+        IndexedRecord dataRecord = new GenericData.Record(mainSchema);
+        dataRecord.put(0, "some name");
+        
+        IndexedRecord outOfBandRecord = new GenericData.Record(outOfBandSchema);
+        dataRecord.put(0, 1);
```
akhabali: this should be outOfBandRecord.put(0,1), No ?  
and you don't use the data in the test so you may just remove that.

gonchar-ivan: yes.
Here should be outOfBandRecord .put(0, 1);
and yes
I'll remove dateRecord.put(), outOfBandRecord.put() because it is not required in this test
Thanks

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/40/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/40/)

gonchar-ivan:
Fail is not related to my change

PR #607, opened 2017-03-22
fix(TDI-38431) : rewrite azure blob components to run in standalone manner

akhabali:

https://jira.talendforge.org/browse/TDI-38431

**What is the new behavior?**
some of azure storage blob don't have main connector and have to work in a standalone manner.
they implement ComponentDriverInitialization interface that provide runAtDriver() method.
This method is used at run time for these kind of component.
The affected components are :
* tAzureStorageContainerCreate
* tAzureStorageContainerExist
* tAzureStorageDelete
* tAzureStorageGet
* tAzureStoragePut

- [ ] Yes
- [x] No

**Other information**:


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/29/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/29/)

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerCreateRuntime.java
```diff
@@ -0,0 +1,122 @@
...
+ * <b>Instances of this class should not be serialized and sent on worker nodes</b>
+ */
+public class AzureStorageContainerCreateRuntime extends AzureStorageContainerRuntime
+        implements ComponentDriverInitialization<ComponentProperties> {
+
+    private static final long serialVersionUID = -8413348199906078372L;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AzureStorageContainerCreateRuntime.class);
+
+    private static final I18nMessages messages = GlobalI18N.getI18nMessageProvider()
```
gonchar-ivan: Just FYI:
Yesterday, I found another way to provide i18n message to the class. 
It is to extend TranslatableImpl. Thus, subclasses will have getI18nMessage() method.
However, this approach is limited. It can be reused only with inheritance. Composition won't work, because path to messages.properties resource will be incorrect. Another limitation is class may extend only 1 class. So, if it already extends something else, it can't extend TranslatableImpl

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerCreateRuntime.java
```diff
@@ -0,0 +1,122 @@
...
+            .getI18nMessages(AzureStorageContainerCreateRuntime.class);
+
+    //
```
gonchar-ivan: unused slashes

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerExistRuntime.java
```diff
@@ -0,0 +1,79 @@
...
+            return false;
+        }
+    }
+
+    private void setReturnValues(RuntimeContainer runtimeContainer, boolean containerExist) {
+        String componentId = runtimeContainer.getCurrentComponentId();
+        String returnContainer = AzureStorageUtils.getStudioNameFromProperty(AzureStorageContainerDefinition.RETURN_CONTAINER);
+        String returnContainerExist = AzureStorageUtils
+                .getStudioNameFromProperty(TAzureStorageContainerExistDefinition.RETURN_CONTAINER_EXIST);
+        //
```
gonchar-ivan: unused

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerRuntime.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private ValidationResult validate() {
+
+        String errorMessage = "";
+        // not empty
+        if (StringUtils.isEmpty(containerName)) {
+            errorMessage = messages.getMessage("error.ContainerEmpty");
+        }
+        // valid characters 0-9 a-z and -
```
gonchar-ivan: There is sense to move such checks to initialize() method.
Both initialize() and validate() checks properties.
But validate() is called on worker node in distributed env and it supposed to check things, which require real connection (check is it possible to establish connection, for instance)
initialize() should be used for all checks, which don't require connection

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStorageGetRuntime.java
```diff
@@ -0,0 +1,156 @@
...
+        implements ComponentDriverInitialization<ComponentProperties> {
+
+    private static final long serialVersionUID = 3439883025371560977L;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AzureStorageGetRuntime.class);
+
+    private static final I18nMessages messages = GlobalI18N.getI18nMessageProvider()
+            .getI18nMessages(AzureStorageGetRuntime.class);
+
+    //
```
gonchar-ivan: unused?

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStorageListReader.java
```diff
@@ -99,13 +99,14 @@ public boolean advance() throws IOException {
...
             currentRecord = new GenericData.Record(properties.schema.schema.getValue());
             currentRecord.put(0, blobs.get(blobIndex).getName());
+            currentBlob = blobs.get(blobIndex);
             return true;
         }
         return false;
     }
 
     @Override
-    public IndexedRecord getCurrent() throws NoSuchElementException {
```
gonchar-ivan: hm... actually it should be thrown.
Check javadoc in interface Reader.getCurrent()
At least if start wasn't called or advance returns false
(this checks that platform/caller correctly uses components)

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStoragePutRuntime.java
```diff
@@ -0,0 +1,161 @@
...
+            fileMap = utils.genAzureObjectList(new File(localFolder), remoteFolder);
+        }
+        for (Map.Entry<String, String> entry : fileMap.entrySet()) {
+            try {
+                CloudBlockBlob blockBlob = blobContainer.getBlockBlobReference(entry.getValue());
+                File source = new File(entry.getKey());
+                FileInputStream stream = new FileInputStream(source);
+                // TODO Any Action ??? if remoteFolder doesn't exist it will fail...
+                blockBlob.upload(stream, source.length());
+                stream.close();
```
gonchar-ivan: maybe move it to finally?

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStoragePutRuntime.java
```diff
@@ -0,0 +1,161 @@
...
+import org.talend.daikon.i18n.GlobalI18N;
+import org.talend.daikon.i18n.I18nMessages;
+import org.talend.daikon.properties.ValidationResult;
+
+import com.microsoft.azure.storage.StorageException;
+import com.microsoft.azure.storage.blob.CloudBlobContainer;
+import com.microsoft.azure.storage.blob.CloudBlockBlob;
+
+/**
+ * upload file to Azure blob storage
```
gonchar-ivan: As I understand it uploads a set of files from some folder, not a single file. Am I right?

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStorageSource.java
```diff
@@ -94,123 +71,10 @@ public BoundedReader createReader(RuntimeContainer container) {
...
-        if (properties instanceof TAzureStoragePutProperties) {
-            TAzureStoragePutProperties props = (TAzureStoragePutProperties) properties;
-            return new AzureStoragePutReader(container, this, props);
-        }
 
         return null;
     }
 
-    @Override
-    public ValidationResult validate(RuntimeContainer container) {
```
gonchar-ivan: Nice!
OOP in action)

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/tazurestoragecontainercreate/TAzureStorageContainerCreateDefinition.java
```diff
@@ -31,16 +34,18 @@ public TAzureStorageContainerCreateDefinition() {
...
-    @SuppressWarnings({"unchecked", "rawtypes"})
+    
     @Override
-    public Class getPropertiesClass() {
-        return TAzureStorageContainerCreateProperties.class;
+    public RuntimeInfo getRuntimeInfo(ExecutionEngine engine, ComponentProperties properties,
+            ConnectorTopology connectorTopology) {
+    	assertConnectorTopologyCompatibility(connectorTopology);
+    	assertEngineCompatibility(engine);
+        return getCommonRuntimeInfo(this.getClass().getClassLoader(), AzureStorageContainerCreateRuntime.class);
```
gonchar-ivan: something wrong with indent

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/tazurestoragecontainercreate/TAzureStorageContainerCreateDefinition.java
```diff
@@ -31,16 +34,18 @@ public TAzureStorageContainerCreateDefinition() {
     public Class<? extends ComponentProperties> getPropertyClass() {
         return TAzureStorageContainerCreateProperties.class;
     }
-
-    @SuppressWarnings({"unchecked", "rawtypes"})
+    
     @Override
-    public Class getPropertiesClass() {
```
gonchar-ivan: hm... seems correct method is getPropertyClass... but I don't know for sure.
Let's keep it like this.
I've created a ticket about it. (We'll fix it in that ticket)
https://jira.talendforge.org/browse/TCOMP-515

components/components-azurestorage/src/test/java/org/talend/components/azurestorage/RuntimeContainerMock.java
```diff
@@ -0,0 +1,49 @@
...
+// ============================================================================
+package org.talend.components.azurestorage;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import org.talend.components.api.container.RuntimeContainer;
+
+public class RuntimeContainerMock implements RuntimeContainer {
```
gonchar-ivan: It duplicates org.talend.components.api.container.DefaultComponentRuntimeContainerImpl
you may extend it and override getCurrentComponentId() method

components/components-azurestorage/src/test/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerCreateRuntimeTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.talend.components.azurestorage.tazurestorageconnection.TAzureStorageConnectionProperties;
+import org.talend.daikon.i18n.GlobalI18N;
+import org.talend.daikon.i18n.I18nMessages;
+import org.talend.daikon.properties.ValidationResult;
+
+public class AzureStorageContainerCreateRuntimeTest {
+
+    public static final String PROP_ = "PROP_";
+
+    private static final I18nMessages messages = GlobalI18N.getI18nMessageProvider()
```
gonchar-ivan: Just thoughts
With such approach we test that returned messages are the same as in .properties resource files.
But we don't test actual message. I'm not sure we should or should not do it.
However, maybe there is some way to set locale before test... or use Junit assume to run specific test only in case of certain locale is set.
We need to think about it. (I don't think we need to test messages for all languages)

components/components-azurestorage/src/test/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerCreateRuntimeTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @After
+    public void dispose() {
+        this.containerCreate = null;
+        properties = null;
+        runtimeContainer = null;
+    }
+
+    @Test
+    public void testNameContainerEmpty() {
```
gonchar-ivan: I suggest to name test methods by following pattern:
test[methodWithIsCalled][testCaseDescription]
e.g.
testInitializeContainerNameEmpty

When you see that testInitializeContainerNameEmpty fails you will jump to directly to initialize() method to fix the bug
If you see testNameContainerEmpty() fails you don't know in which method the problem appears. Thus, you need to read test case to find out what is called inside

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerRuntime.java
akhabali: this is already used in the initialize() method.
The validate() method here is just a private method that encapsulate the validation logic and it's called in the initialize method whit a return return validate().
I will remove this method and move the validation logic directly to the initialize method cause it's not huge.

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStoragePutRuntime.java
akhabali: Yes, i will update the java doc

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerRuntime.java
gonchar-ivan: yes. It's ok.
Sorry for my inattention

components/components-azurestorage/src/main/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerCreateRuntime.java
akhabali: Yes, the usage of inheritance is limited and may limit the conception of an object. 
but implementing an interface may be more suitable for this kind of things. Like the Cloneable interface in the JDK that provide clone abilities to an object.We may use the same concept to provide translation abilities to a class.

components/components-azurestorage/src/test/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerCreateRuntimeTest.java
akhabali: The test and the I18nMessages use the machine default Locale. so it's OK. 
Sure we are not testing returned messages in all languages but i think that it's sufficient.

What do you mean by 

> we don't test actual message.

components/components-azurestorage/src/test/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerCreateRuntimeTest.java
akhabali: I like this approach, but did we have some naming convention rules some where?

components/components-azurestorage/src/test/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerCreateRuntimeTest.java
gonchar-ivan: I mean in .properties you may have some incorrect or unclear message.
e.g. "Error!"
However test will pass.
When you change message in .properties file to
"Property xxx wasn't set"
Test will pass again.

So, by this comment I wanted to raise a question: How should we test messages?
On the one hand: 
1. We don't want to create tests for each language
2. Seems we can't test only some particular language, for instance English, because i18n feature depends on locale. So, on one computer (with English locale) tests may pass, and they may fail on another (with another locale)
Or is there something we should set to guarantee that we test English messages?
On the other hand:
1. If we don't test actual messages, somebody may break message and we won't know about it from unit-tests (We will know about it only from manual checking or from user bug)

components/components-azurestorage/src/test/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerCreateRuntimeTest.java
gonchar-ivan: No we haven't. I am creating a page about it

components/components-azurestorage/src/test/java/org/talend/components/azurestorage/blob/runtime/AzureStorageContainerCreateRuntimeTest.java
akhabali: Here what matter for me, is that the message returned by the validation method match the message that  i retrieve using the message property key.
For example
a validate method that return a validation message 
```
public ValidationResult initialize(Properties properties){
    return  messages.getMessage("error.ContainerEmpty");
}
```
The test will be 
`assertEquals(messages.getMessage("error.ContainerEmpty"), initialize(props))`
i don't care about the message text and the used locale. what i need to check is that the same text is returned under a key for a use case.
You see what i mean.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/81/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/81/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/88/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/88/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/85/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/85/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/2/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/2/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/7/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/7/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/9/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/9/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/13/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/13/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/14/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/14/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/8/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/8/)

PR #608, opened 2017-03-22
Feature/tdi 38101 marketo dynamic schema

undx:

https://jira.talendforge.org/browse/TDI-38101

**What is the new behavior?**

_CustomObjects_ are managed in 
* `tMarketoInput` : actions available are `describe`, `list` and `get`.
* `tMarketoOutput` : actions available are `syncCustomObject` and `deleteCustomObject`

_Dynamic Schema support_ supported on Leads (getLead, getMultipleLeads) and CustomObjects (get).

- [ ] Yes
- [X] No

**Other information**:


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/30/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/30/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/31/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/31/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/33/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/33/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/34/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/34/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/35/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/35/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/36/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/36/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/41/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/41/)

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoInputReader.java
```diff
@@ -49,98 +58,112 @@
...
+    private transient static final Logger LOG = LoggerFactory.getLogger(MarketoInputReader.class);
+
+    private static final I18nMessages messages = GlobalI18N.getI18nMessageProvider().getI18nMessages(MarketoInputReader.class);
+
     public MarketoInputReader(RuntimeContainer adaptor, MarketoSource source, TMarketoInputProperties properties) {
         super(source);
         this.adaptor = adaptor;
         this.source = source;
         this.properties = properties;
+        // check if we've a dynamic schema...
```
jzhao-talend: `AvroUtils` provide a method `AvroUtils.isIncludeAllFields(...)` to check dynamic schema

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSource.java
```diff
@@ -66,6 +67,16 @@ public ValidationResult validate(RuntimeContainer container) {
         if (properties instanceof TMarketoInputProperties) {
             TMarketoInputProperties p = (TMarketoInputProperties) properties;
             boolean useSOAP = properties.getConnectionProperties().apiMode.getValue().equals(APIMode.SOAP);
+            // Validate dynamic schema if needed
+            Boolean isDynamic = p.schemaInput.schema.getValue()
+                    .getProp(DiSchemaConstants.TALEND6_DYNAMIC_COLUMN_POSITION) != null;
+            if (useSOAP) { // no dynamic schema for SOAP !
```
jzhao-talend: As I understand that dynamic only for `REST` mode right? Is it possible to get dynamic setting by retrieve query result which like what we done in some components (like tSalesforceInput manual query mode)
This implement is done after discuss with Pierre ?

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
```diff
@@ -132,6 +136,84 @@ public Schema getEndpointSchema(RuntimeContainer container, String schemaName) t
...
+     *
+     * @param the ObjectName to get schema. If blank, assumes it's for Lead fields. Otherwise ObjectName should be the
+     * CustomObject API's name.
+     * 
+     * @return the schema for the given CustomObject or all Lead fields.
+     * 
+     */
+    public Schema getDynamicSchema(String objectName, Schema design) throws IOException {
+        // check if dynamic
+        String strdynpos = design.getProp(DiSchemaConstants.TALEND6_DYNAMIC_COLUMN_POSITION);
```
jzhao-talend: Use `AvroUtils.isIncludeAllFields(...)` , need to check other place which use same kind of code.

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
```diff
@@ -132,6 +136,84 @@ public Schema getEndpointSchema(RuntimeContainer container, String schemaName) t
...
+        if (strdynpos == null)
+            return design;
+        // check object managed
+
+        // split existing fields according Schema dynamic column's position.
+        int dynPos = Integer.parseInt(strdynpos);
+        List<Field> beforeFields = new ArrayList<>();
+        List<Field> afterFields = new ArrayList<>();
+        List<String> existingFieldNames = new ArrayList<>();
+        for (Field f : design.getFields()) {
```
jzhao-talend: I have some question about this part of code:
Are all designed fields exists in `objectFields` which fetched from server ?
If yes, I remember that we have common code in framework to handle dynamic fields with design field.
https://github.com/Talend/daikon/tree/master/daikon/src/main/java/org/talend/daikon/di
we only need get all fields list.

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSource.java
undx: I've implemented it in CustomObject _get_ operation for the tMarketoInput. I've had a _Fetch schema_ button which brings back the custom object fields.
In the _describe_ and _list_ operations you cannot have a custom schema, it's a fixed one.

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoInputReader.java
undx: Ok, I'll replace this code by calling this method.

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
undx: Sorry, I don't see where I can get the full merged ordered schema in the code mentionned...

Is it preserving the order of the user's defined schema fields ?

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
undx: I've replaced the code where I needed a boolean value. Here, I just need the index position, so it's OK.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/61/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/61/)

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
jzhao-talend: It should done automaticly. It think you can try in you side.
It is done in generic part.
https://github.com/Talend/tdi-studio-se/blob/master/main/plugins/org.talend.designer.codegen/jet_stub/generic/component_util_indexedrecord_to_rowstruct.javajet#L108

https://github.com/Talend/daikon/blob/master/daikon/src/main/java/org/talend/daikon/di/DiOutgoingSchemaEnforcer.java#L149

It would create a map of correspondence between design and runtime fields, when first record is wrapped

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoInputReader.java
```diff
@@ -49,98 +58,111 @@
...
         }
-
         return startable;
     }
 
     @Override
     public boolean advance() throws IOException {
         recordIndex++;
-        if (recordIndex < records.size()) {
+        if (recordIndex < records.size())
```
jzhao-talend: [code style] we prefer to use that before current change.

gonchar-ivan: https://github.com/Talend/tools/blob/master/tools-java-formatter/src/main/resources/talend_java_eclipse_formatter.xml

This is our company-wide java formatting rules

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
gonchar-ivan: It is not allowed to use DiSchemaConstants class. It is planned to move all Di* classes from daikon (because this functionality is not common, but Studio (aka DI) specific)
However, I agree here with you. I met the same problem. Dynamic column position is required in component to build runtime/dynamic schema correctly.

So, I see 2 solutions:
1. Move DiSchemaConstants.TALEND6_DYNAMIC_COLUMN_POSITION to daikon SchemaConstants 
(but architect don't want to see anything "dynamic" in daikon, because dynamic feature is DI-specific)
2. Modify SchemaConstants.INCLUDE_ALL_FIELDS to store dynamic column position
(Seems change on Studio side is required to set dynamic column position to INCLUDE_ALL_FIELDS property)

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
gonchar-ivan: Sorry, seems you are talking about different things:

1. If design schema contains dynamic column, then component should create actual (contains all entity fields) runtime schema by itself. So, I think Manu is doing correctly here.

2. There are 2 things, which are used in generated code:
a. codeVarIndexedRecordAdapter. If you check current component repo wiki:
https://github.com/Talend/components/wiki/6.-Component-Runtime#emitting-data-from-input-components
you'll see that it is allowed to emit data-store specific objects. Actually it is forbidden now... but it requires a lot of changes, so it will be done a little later. I mentioned about it in components chat and create Jira to do required job.
So, codeVarIndexedRecordAdapter is/was used to convert data-store specific object to IndexedRecord. This thing is never used correctly, because all our components emit IndexedRecord.
b. codeVarSchemaEnforcer. It converts Avro IndexedRecord to Di record. This is required in Studio for compatibility. Other products which use components won't do this. They will directly pass IndexedRecord between components.
But in Studio TCOMP component may be used with old component in the same job. That's why Studio always converts IndexedRecord to DI record (rowXstruct) after Input component.

And there is no functionality, which constructs dynamic runtime schema in codegen.
(Or I don't know something. Sorry, if I'm wrong)

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
gonchar-ivan: please, create a jira about it

undx: Ok, I'll create one in daikon.
For now, is there a workaround for that ? How did you do ?

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoInputReader.java
undx: You mean to put braces for only one instruction in conditionals 
`if (recordIndex < records.size())
      return true;
`
=>
`if (recordIndex < records.size()) {
      return true;
}`

gonchar-ivan: yes
the same with for/while loops

As I know such convention is widely used

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
undx: JIRA issue created (https://jira.talendforge.org/browse/TDKN-154).

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/86/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/86/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/87/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/87/)

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
jzhao-talend: @gonchar-ivan 
```
 // will fetch fields...
MarketoRESTClient client = (MarketoRESTClient) getClientService(null);
if (StringUtils.isEmpty(objectName)) {
     objectFields = client.getAllLeadFields();
} else {
      objectFields = getSchemaFieldsList(getEndpointSchema(null, objectName));
}
```
Because I found there is some code can get all fields for the Object. 
So I think there only need to included all fields to runtime schema. we no need build schema based on the design schema.

For example:
we have design schema:
```
col1
col3
dynamic
col6
```
if have a way (like: `getAllLeadFields`) get all fields list:
```
col1
col2
col3
col4
col5
col6
```
we only need build runtime schema with those fields directly. and no need keep the field order with design schema.
Then when we use this compoent pass record to other component, it would create a map of correspondence between design and runtime fields. 
If I am wrong about this, you still get my means. I think we can discuss in marketo channel.

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSinkSchemaProvider.java
```diff
@@ -12,15 +12,15 @@
...
 package org.talend.components.marketo.runtime;
 
+import java.io.IOException;
+
 import org.apache.avro.Schema;
 import org.talend.components.api.component.runtime.SourceOrSink;
-import org.talend.components.api.properties.ComponentProperties;
 
 /**
  * Created by undx on 25/01/2017.
```
jzhao-talend: Clean this kind of comments

components/components-marketo/src/main/java/org/talend/components/marketo/runtime/MarketoSourceOrSink.java
gonchar-ivan: Oh... yes
You are right! I'm sorry.

I just had a special case. I was working on input-component-archetype (which is simple version of fileDelimited component).
For this component only user can specify correct type of some column. Because in file everything is String and component can't guess whether "5.4" is a String or a Double.
So, when component creates Runtime schema it sets String type for each column.
Consider following example of a file:
"25;Ivan;true;123.345"
In case of following design schema:
1. col1 Integer // user knows that first column is Integer
2. dynamic Dynamic

Component should copy all non-dynamic column from design schema to runtime schema (to save their types). Also it should know dynamic position to insert all rest columns on that position.

But if component can retrieve column types from a) API  or b) first data record, then it doesn't require to care about dynamic position.

jzhao-talend: yes. this special case should be considered.
@undx  could you confirm with Ivan and my comments?
If you have special case which Ivan met, then you can keep current way to generate dynamic runtime schema.
else you need change current code to return the runtime schema just with fields which you fetch from server side.
then maybe dynamic position would no need here.

undx: @jzhao-talend, I'm not in that special case with the leads schema. I have to check with the custom objects schema. I'll experiment if the result schema is well ordered. In that case, I'll remove the dynamic position relative code.

undx: OK, it works in the studio without having to manage dynamic position. Cool !
I rework the code and push.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/4/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/4/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/5/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/5/)

PR #612, opened 2017-03-24
TCOMP-92: Need to define the palette type for each component

ycbai:

**What is the new behavior?**

- [ ] Yes
- [x] No

**Other information**:


build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/49/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/49/)

core/components-api/src/main/java/org/talend/components/api/component/SupportedProduct.java
```diff
@@ -0,0 +1,26 @@
+package org.talend.components.api.component;
+
+/**
+ * Use to define product type which is supported by component.
+ */
+public enum SupportedProduct {
```
gonchar-ivan: What if we will need to add more products in future?
Then we have to add one more enum value. Won't it break client code?
(Its just to think, not a concern)
(I think it shouldn't break if client code doesn't handle returned value in switch-case expression)

core/components-api/src/main/java/org/talend/components/api/component/ComponentDefinition.java
```diff
@@ -162,4 +163,11 @@
      */
     Set<ConnectorTopology> getSupportedConnectorTopologies();
 
+    /**
+     * This method return which products the component supports.
+     *
+     * @return all supported product types.
+     */
+    EnumSet<SupportedProduct> getSupportedProducts();
```
gonchar-ivan: What about wizards? Do we need to such logic for wizard definitions?

Maybe we need to declare this method higher in hierarchy... 
However, me may do it later, because it will be safe change

sgandon: that is a good point here.

core/components-api/src/main/java/org/talend/components/api/component/SupportedProduct.java
sgandon: I am not sure this is an issue, the clients will just check that the `getSupportedProducts()` contain their own product. 
I also feel that using a simple String constant would be more flexible but I can explain why.

ycbai: @gonchar-ivan sure you need to update this enum if add more products after. But it will not break client code because as @sgandon said the client only concern if this product it supports or not.

ycbai: @sgandon why String constant is more flexible then enum?  Could you explain it in detail? Thanks.

core/components-api/src/main/java/org/talend/components/api/component/ComponentDefinition.java
ycbai: @gonchar-ivan @sgandon There is no need to add it to wizard definitions because wizards are not grouped by palette or such.

core/components-api/src/main/java/org/talend/components/api/component/SupportedProduct.java
sgandon: that is the thing it is just a feeling related to the fixed enum type which adds more constraint to the type. 
If someone make a component their given product they will have to update this enum where as if we only use string, they can create a new string for a new product without having to change the components API. This is why it seems more flexible. 
It is like the widget name for the component layout, you can use a new widget without having to modify daikon for it because this is only a String label.

ycbai: OK, I see. Thanks!

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/77/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/77/)

nrousseau:
Note that there is actually one break of API for other products, since the ComponentDefinition has been modified.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/15/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/15/)

ycbai:
I dont know how this PR break those junits?

sgandon:
Please merge, this issues comes from jenkins not being configured with the right credentials on slave nodes.

PR #614, opened 2017-03-24
bugfix(TDI-38263): fix NPE in Guess Schema feature

TotalP:
NPE when run the job after Guess Schema feature used. There are no default date pattern, length and precision props.   

**What is the new behavior?**
Added default date pattern, length and precision props

- [ ] Yes
- [x] No

**Other information**:
https://jira.talendforge.org/browse/TDI-38263

### DIFF
--- a/components/components-salesforce/components-salesforce-runtime/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
+++ b/components/components-salesforce/components-salesforce-runtime/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
@@ -19,7 +19,9 @@ import java.io.IOException;
 import java.net.Authenticator;
 import java.net.PasswordAuthentication;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 
 import javax.xml.namespace.QName;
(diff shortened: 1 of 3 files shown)


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/52/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/52/)

components/components-salesforce/components-salesforce-runtime/src/main/java/org/talend/components/salesforce/runtime/SalesforceSourceOrSink.java
```diff
@@ -491,18 +492,64 @@ public Schema guessSchema(String soqlQuery) throws IOException {
...
+                    schema.getField(fieldDescription.getSimpleName())
+                            .addProp(entry.getKey(), String.valueOf(SalesforceConstants.DEFAULT_LENGTH));
+                }
+                if (entry.getKey().equals(SchemaConstants.TALEND_COLUMN_PRECISION)) {
+                    schema.getField(fieldDescription.getSimpleName())
+                            .addProp(entry.getKey(), String.valueOf(SalesforceConstants.DEFAULT_PRECISION));
+                }
+            }
+        }
+        return schema;
```
chmyga: You always set the default values here. What if those properties are already set(retrieved from salesforce DescribeSObject)? We will reset those properties to default values.

PR #628, opened 2017-03-31
fix(TDI-38447): Fixed dependencies.

iistomin-gl:
* The dependencies.txt includes incorrect set of dependecies

**What is the new behavior?**
* The dependencies.txt includes dependecies that are required for components

- [X] Bugfix
- [ ] Feature

- [ ] Yes
- [X] No

**Other information**:
JIRA issue: https://jira.talendforge.org/browse/TDI-38447


components/components-netsuite/netsuite-runtime/pom.xml
```diff
@@ -61,12 +55,15 @@
             <artifactId>commons-beanutils</artifactId>
             <version>1.8.3</version>
         </dependency>
-
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>18.0</version>
```
gonchar-ivan: Recently, I have changed guava dependency in parent pom to 19.0 to fix dependency problem
So, if this change was made due to problem in Studio, it might already disappear

gonchar-ivan: Another thing: do you use it directly in your component?
If not better to declare it in dependencyManagement block to force version of transitive dependency.
I use mvn dependency:analyze goal to check whether I declared dependencies correctly

iistomin-gl: Yes, guava is used directly in netsuite-runtime.

gonchar-ivan: ok.
Then it should be declared in runtime pom.xml
but you may remove version.
Version will be controlled by parent pom.

(Actually I started to work on maven poms. I want to "separate" sdk from componens and use parent BOM also. I will show first results soon)

sgandon: Are you sure you need guava in the definition module ?

components/components-netsuite/netsuite-runtime/pom.xml
```diff
@@ -61,12 +52,14 @@
             <artifactId>commons-beanutils</artifactId>
             <version>1.8.3</version>
         </dependency>
-
+        <dependency>
```
sgandon: are you sure you need guava and javapoet in the definition module ?
The definition should not contain any domain specific dependencies of your components, only runtime classes should have those.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/86/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/86/)

PR #636, opened 2017-04-04
fix(TDI-38549): Implemented bulk processing for output component

iistomin-gl:
- For each written record NetSuite output component calls NetSuite web service (add/update/upsert/delete)

**What is the new behavior?**
- NetSuite output component processes records in a bulk and calls NetSuite web service bulk operation (addList/updateList/upsertList/deleteList)

- [X] Docs have been added / updated (for bug fixes / features) ?

- [X] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)

- [ ] Yes
- [X] No

**Other information**:
JIRA issue: https://jira.talendforge.org/browse/TDI-38549


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/108/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/108/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/110/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/110/)

components/components-netsuite/netsuite-runtime/src/main/java/org/talend/components/netsuite/output/NetSuiteOutputWriter.java
```diff
@@ -88,10 +110,28 @@ public void open(String uId) throws IOException {
...
+            switch (action) {
+            case ADD:
+                bulkWriteProcessor = new BulkAddProcessor<>();
+                break;
+            case UPDATE:
+                bulkWriteProcessor = new BulkUpdateProcessor<>();
+                break;
+            case UPSERT:
+                useNativeUpsert = Boolean.TRUE.equals(writeOperation.getProperties()
+                        .module.useNativeUpsert.getValue());
```
gonchar-ivan: I'm not sure Boolean.TRUE.equals is required here
What about just use:
writeOperation.getProperties().module.useNativeUpsert.getValue()
?

components/components-netsuite/netsuite-runtime/src/main/java/org/talend/components/netsuite/output/NetSuiteOutputWriter.java
```diff
@@ -88,10 +110,28 @@ public void open(String uId) throws IOException {
             transducer = new NsObjectOutputTransducer(clientService, typeDesc.getTypeName());
             transducer.setMetaDataSource(metaDataSource);
 
-            if (action == OutputAction.UPDATE || action == OutputAction.UPSERT) {
-                transducer.setRecordSource(new NsObjectOutputTransducer.DefaultRecordSource(clientService));
-            } else if (action == OutputAction.DELETE) {
+            switch (action) {
```
gonchar-ivan: I suggest to move whole switch to private method:
initializeBulkProcessor
or
createBulkProcessor

components/components-netsuite/netsuite-runtime/src/main/java/org/talend/components/netsuite/output/NetSuiteOutputWriter.java
```diff
@@ -42,14 +48,30 @@
...
     protected boolean exceptionForErrors = true;
 
+    /**
+     * Specifies writer should use native <code>upsert</code> operation or
+     * should use <code>add</code> and <code>update</code> for new and existing records correspondingly.
+     *
+     * <p>Use of native <code>upsert</code> operation disabled by default, it can be changed by
+     * {@link NetSuiteOutputModuleProperties#useNativeUpsert} property.
+     */
+    protected boolean useNativeUpsert = false;
```
gonchar-ivan: 1. This value is not changed during Writer object lifecycle
2. It duplicates value from Properties

I suggest you:
1. Create bulkWriteProcessor in NetSuiteWriteOperation.createWriter() method
NetSuiteWriteOperation contains instance of Properties. Thus, it may easily access to `action` and `useNativeUpsert`, which are required for bulkWriteProcessor creation
2. Then inject (pass as argument) bulkWriteProcessor to NetSuiteOutputWriter constructor.
3. Thus, you may remove `action` and `useNativeUpsert` fields from Writer. In Writer these fields are static, also they duplicates values from Properties.

components/components-netsuite/netsuite-runtime/src/main/java/org/talend/components/netsuite/output/NetSuiteOutputWriter.java
```diff
@@ -139,8 +162,238 @@ public Result close() throws IOException {
...
+     *     <li>Add - {@link BulkAddProcessor}</li>
+     *     <li>Update - {@link BulkUpdateProcessor}</li>
+     *     <li>Upsert - {@link BulkNativeUpsertProcessor}, {@link BulkCustomUpsertProcessor}</li>
+     *     <li>Delete - {@link BulkDeleteProcessor}</li>
+     * </ul>
+     *
+     * @param <T> type of NetSuite objects that are passed to {@link NetSuiteClientService}
+     * @param <RefT> type of NetSuite reference objects
+     */
+    abstract class BulkWriteProcessor<T, RefT> {
```
gonchar-ivan: As for me a lot of inner classes inside one outer class complicates code.
Short classes are easier to read/understand/maintain
So, I prefer to not use inner classes if it is possible.
But it is up to you. Because I don't know which approach is correct. Also I would be glad if you tell me some cons and pros of your approach.

gonchar-ivan: Another thing here.
Isn't it easier to extend Writer class?
You may have parent abstract Writer class with all common functionality and several implementations:
UpdateWriter, DeleteWriter, UpsertWriter etc, which overrides doWrite() method.

Motivation:
Here you introduce new abstraction - BulkWriteProcessor. So, some other developer should spend some time to understand what it is. However, other developer already knows what is Writer abstraction.
Also, BulkWriteProcessor duplicates Writer logic. It does the same actions as Writer does.
What do you think about it?

components/components-netsuite/netsuite-runtime/src/main/java/org/talend/components/netsuite/output/NetSuiteOutputWriter.java
```diff
@@ -139,8 +162,238 @@ public Result close() throws IOException {
...
+
+        private int batchSize = DEFAULT_BATCH_SIZE;
+
+        // Holds accumulated IndexedRecords for a current batch
+        private List<IndexedRecord> inputRecordList = new ArrayList<>();
+
+        int getBatchSize() {
```
gonchar-ivan: is this method used?
I haven't found its usages.

I suggest to remove getBatchSize() and setBatchSize()
You may add them later easily if you need.
Generally, I have a rule to not add something for future in case I see that I can add it easily, when I need.
(If it is not easy to add it later, then I agree - it should be added at the beginning)

components/components-netsuite/netsuite-runtime/src/main/java/org/talend/components/netsuite/output/NetSuiteOutputWriter.java
```diff
@@ -28,12 +31,15 @@
...
 import org.talend.components.netsuite.client.model.TypeDesc;
+import org.talend.components.netsuite.client.model.beans.Beans;
 
 /**
- * TODO Implement bulk Add/Update/Upsert/Delete
+ *
  */
 public class NetSuiteOutputWriter implements WriterWithFeedback<Result, IndexedRecord, IndexedRecord> {
 
+    public static final int DEFAULT_BATCH_SIZE = 100;
```
gonchar-ivan: I suggest to move this constant to Properties class, where batchSize property is declared.
(If it is not declared in Properties, then it is ok to leave it here. However, consider to add such property. Most components allow user to specify batch size)

components/components-netsuite/netsuite-runtime/src/main/java/org/talend/components/netsuite/output/NetSuiteOutputWriter.java
```diff
@@ -42,14 +48,30 @@
...
     protected TypeDesc typeDesc;
+
     protected NsObjectOutputTransducer transducer;
 
+    /**
+     * Performs bulk processing of written NetSuite objects.
+     */
+    protected BulkWriteProcessor<?, ?> bulkWriteProcessor;
+
     protected int dataCount = 0;
```
gonchar-ivan: I think you may replace this field with Result instance field.

components/components-netsuite/netsuite-runtime_2016_2/src/test/java/org/talend/components/netsuite/v2016_2/output/NetSuiteOutputWriterTest.java
```diff
@@ -100,27 +102,35 @@ public void testUpdate() throws Exception {
...
-                recordRef.setType(RecordType.OPPORTUNITY);
-
-                updatedRecordList.add(record);
-
-                UpdateResponse response = new UpdateResponse();
-                WriteResponse writeResponse = new WriteResponse();
-                writeResponse.setStatus(createSuccessStatus());
-                writeResponse.setBaseRef(recordRef);
-                response.setWriteResponse(writeResponse);
+        when(port.updateList(any(UpdateListRequest.class))).then(new Answer<UpdateListResponse>() {
```
gonchar-ivan: Seems, this is to much effort for creating mocks.
This was one of the Francis arguments against mocks. If it is easier/quicker to create a test with real connection, then better to do it rather than inventing a mock.

components/components-netsuite/netsuite-runtime_2014_2/src/main/java/org/talend/components/netsuite/v2014_2/client/NetSuiteClientServiceImpl.java
```diff
@@ -174,6 +175,24 @@ public NetSuiteClientServiceImpl() {
     }
 
     @Override
+    public <RecT, RefT> List<NsReadResponse<RecT>> getList(final List<RefT> refs) throws NetSuiteException {
+        if (refs == null) {
+            return Collections.emptyList();
+        }
+        return execute(new PortOperation<List<NsReadResponse<RecT>>, NetSuitePortType>() {
```
dmytro-sylaiev: It would be great to rewrite this return, quite hard to understand what is going on.

components/components-netsuite/netsuite-runtime_2016_2/src/test/java/org/talend/components/netsuite/v2016_2/output/NetSuiteOutputWriterTest.java
iistomin-gl: Real integration tests are currently disabled, so mock based tests are really useful for me now. I don't invent anything, I'm just testing how components use NetSuite API.

components/components-netsuite/netsuite-definition/src/test/java/org/talend/components/netsuite/NetSuiteVersionTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+package org.talend.components.netsuite;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ *
+ */
```
iistomin-gl: I don't think that comment will be useful. Name of class and test method gives information what this class tests.

components/components-netsuite/netsuite-runtime_2014_2/src/main/java/org/talend/components/netsuite/v2014_2/client/NetSuiteClientServiceImpl.java
iistomin-gl: I don't clearly understand which part of code looks difficult to read.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/115/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/115/)

components/components-netsuite/netsuite-runtime/src/main/java/org/talend/components/netsuite/output/NetSuiteOutputWriter.java
```diff
@@ -17,43 +17,74 @@
...
+    public boolean isExceptionForErrors() {
+        return exceptionForErrors;
+    }
+
+    public void setExceptionForErrors(boolean exceptionForErrors) {
+        this.exceptionForErrors = exceptionForErrors;
+    }
+
+    public int getBatchSize() {
+        return batchSize;
```
gonchar-ivan: just note you may do following here:
return writeOperation.getProperties().batchSize.getValue();

and replace batchSize field usage with a call to getBatchSize()
(Thus, you won't duplicate this value. Now it appears in 2 places: NetSuiteOutputProperties and this Writer class)

components/components-netsuite/netsuite-runtime_2016_2/src/test/java/org/talend/components/netsuite/v2016_2/output/NetSuiteOutputWriterTest.java
gonchar-ivan: ok. Agree with you.
"Invent" was wrong word here. I mean to "write" mock.

components/components-netsuite/netsuite-definition/src/test/java/org/talend/components/netsuite/NetSuiteVersionTest.java
gonchar-ivan: Agree with Ihor

components/components-netsuite/netsuite-runtime_2014_2/src/main/java/org/talend/components/netsuite/v2014_2/client/NetSuiteClientServiceImpl.java
gonchar-ivan: I guess "return" expression.
I tend to agree with Dmytro, but I don't know how it may be simplified.
Also seems it is a matter of habit. If you often meet (or write by yourself) such expressions, it won't be hard for you to understand them.
I think it may be left without changes.

PR #654, opened 2017-04-11
fix(TFD-865): Prevent committing Hadoop output with no results.

RyanSkraba:

Beam attempts to commit a "job" when a Hadoop OutputFormat is used.  Every shard must return the file name that it created, but our writers only create a file when there is data to be written (lazy writing).  When no writers commit results, the job should not be committed.

Beam already has implemented this behaviour when there are NO shards (as opposed to no valid shards).

**What is the new behavior?**

Shards that have not produced results return `null` as the file name.  These are removed from the aggregated results, and the job is committed only if at least one of the writers produced results.

If no writers produced results, the same logic is used as when there are no shards.

- [X] Bugfix
- [ ] Feature

- [ ] Yes
- [X] No

**Other information**:


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/140/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/140/)

PR #727, opened 2017-05-11
fix(unittests):fix 3 unit tests in service spi

sgandon:
3 unit test are failing on master cause they are not finding the right version for the uri : `mvn:org.talend.components/multiple-runtime-comp/`
this is a know issue from pax-maven-url.

**What is the new behavior?**
Test are ok now as they are requesting a fixed version
`mvn:org.talend.components/multiple-runtime-comp/0.18.0`

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/337/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/337/)

services/components-api-service-spi-test/pom.xml
```diff
@@ -22,26 +22,26 @@
 		<dependency>
 			<groupId>org.talend.components</groupId>
 			<artifactId>localio-definition</artifactId>
-			<version>${project.version}</version>
+			<version>0.18.0</version>
```
akhabali: can you use the actual dev version 0.19.0-SNAPSHOT ?

sgandon: there is not reason to use the current version and more over this will cause pb cause the version is also in the code and you don't want to change the code everytime you change the version.
For those tests we dont need any update of those artifacts.

gonchar-ivan: Current release process may break this version. mvn release plugin may try to upgrade this version.
(It did so for zeLib. I need to check)

Another possible solution here could be to expose project.version as some system property with maven-failsafe-plugin and use that property in tests

PR #739, opened 2017-05-18
bugfix(TDI-38859) : fixed doubled last output on flows

undx:

https://jira.talendforge.org/browse/TDI-38859

**What is the new behavior?**

- [X] Bugfix
- [ ] Feature

- [ ] Yes
- [X] No

**Other information**:


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/373/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/373/)

components/components-marketo/src/main/java/org/talend/components/marketo/tmarketooutput/TMarketoOutputDefinition.java
```diff
@@ -58,11 +58,6 @@ public boolean isConditionalInputs() {
     }
 
     @Override
-    public boolean isRejectAfterClose() {
```
jzhao-talend: @undx 
I remember this function use to handle  batch mode output reject.
I didn't did you explanation about why this caused the duplicate records.
If we don't support reject for batch mode (```syncMuiltipleLeads```) , this modification is ok.

undx: Yes and it seems to double the last record in outputs when set.

jzhao-talend: So you should remove the duplicate set record, rather than just disable reject for batch mode.
And I remember that tMarketoListOperation not support batch mode. That was removed is ok.

undx: Batch mode is not use in this issue. The issue comes in 1-1 mode, seems that this parameter in definitions makes that issue on flows (I'll make a recheck).
tMarketoListOperation supports also batch mode (add or remove many leads option).

jzhao-talend: Yes. This issue is for 1-1 mode, but it caused by handle reject.
I means that if we don't support reject for handle reject. your current modification is ok.
If we need to support reject for handle reject, you should fix the issue in this pull request.

undx: I confirm that `isRejectAfterClose` duplicates the last row in 1-1 mode. Returning `false` fixes the issue.
I think that I will re-enable outputs in batch mode when a decision will be made on the related [issue](https://jira.talendforge.org/browse/TDI-38610).

PR #763, opened 2017-05-23
feat(TDKN-127): add format properties provider

chmyga:
* Add FormatPropertiesProvider

* Add DefinitionRegistry injection for properties

* Add post deserialization process for definition registry injection

* Add tests

**What is the new behavior?**
Add FormatPropertiesProvider which gets possible format definitions/properties from definition registry.

- [ ] Bugfix
- [x] Feature

- [ ] Yes
- [x] No

**Other information**:
https://jira.talendforge.org/browse/TDKN-127


build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/411/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/411/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/413/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/413/)

core/components-common/src/main/java/org/talend/components/common/format/FormatPropertiesProvider.java
```diff
@@ -0,0 +1,157 @@
...
+    private void updateProperties() {
+        String selectedFileFormatDefinitionStr = format.getValue();
+        if (!StringUtils.isEmpty(selectedFileFormatDefinitionStr)
+                && !StringUtils.equals(previousFormatValue, selectedFileFormatDefinitionStr)) {
+            V props = referenceMemento.get(selectedFileFormatDefinitionStr);
+            if (props == null) {
+                Definition<V> fileFormatDefinition = getDefinitionRegistry().getDefinitionsMapByType(getDefinitionClass())
+                        .get(selectedFileFormatDefinitionStr);
+                props = getDefinitionRegistry().createProperties(fileFormatDefinition, FORMAT_PROPERTIES_NAME);
+                referenceMemento.put(selectedFileFormatDefinitionStr, props);
```
sgandon: Do we really need this memento, it seems to be a cache but I wonder if this does not make things more complicated for nothing.

core/components-common/src/main/java/org/talend/components/common/format/FormatPropertiesProvider.java
```diff
@@ -0,0 +1,157 @@
...
+            }
+            formatProperties = props;
+        } else if (!StringUtils.isEmpty(selectedFileFormatDefinitionStr)
+                && StringUtils.equals(previousFormatValue, selectedFileFormatDefinitionStr)) {
+            // After deserialization we already have formatProperties, but we need to put them to referenceMemento
+            referenceMemento.put(selectedFileFormatDefinitionStr, (V) formatProperties);
+        } else {
+            // We can not leave properties field null, otherwise we will get an exception during properties
+            // initialization. If no definitions were found, we don't need any properties.
+            formatProperties = PropertiesImpl.createNewInstance(PropertiesImpl.class, FORMAT_PROPERTIES_NAME);
```
sgandon: here I think you can avoid doing this by overriding `org.talend.daikon.properties.PropertiesImpl.acceptUninitializedField(Field)`
see https://github.com/Talend/daikon/blob/sgandon/CQRS_POC/daikon/src/main/java/org/talend/daikon/properties/ReferenceProperties.java#L65

core/components-common/src/main/java/org/talend/components/common/format/FormatPropertiesProvider.java
```diff
@@ -0,0 +1,157 @@
...
+    private static final long serialVersionUID = 7293969314564405598L;
+
+    private static final String FORMAT_PROPERTIES_NAME = "formatProperties";
+
+    @Inject
+    public transient DefinitionRegistryService definitionRegistry;
+
+    public final Property<String> format = PropertyFactory.newString("format");
+
+    public Properties formatProperties;
```
sgandon: this should not be of type V ?

core/components-common/src/main/java/org/talend/components/common/format/FormatPropertiesProvider.java
chmyga: yes! that's what we need here! thank you!

core/components-common/src/main/java/org/talend/components/common/format/FormatPropertiesProvider.java
chmyga: If we change the selection of format property, previous properties would be lost here. So if we select some other properties(even not on purpose) and then switch back to previous properties, we will loose everything we already set.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/462/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/462/)

core/components-api/src/main/java/org/talend/components/api/service/ComponentService.java
```diff
@@ -208,4 +208,6 @@
     Set<? extends Connector> getAvailableConnectors(ComponentProperties componentProperties,
             Set<? extends Connector> connectedConnetor, boolean isOuput);
 
+    <T extends Properties> void postDeserialize(T props);
```
sgandon: I don't understand the reason for this method since it is never called, why would we need that ?

sgandon: Ha I think I get it this is for the Studio once they have deseriazed a properties they should call this method.
But I think this should be done in daikon though.
probably change the methods in `org.talend.daikon.properties.Properties.Helper`or add new methods that would take the DefinitionRegistry as a paramter to be able to do the injection there. And also move the injection code in daikon, it may be reused.

chmyga: I will try to implement that.

chmyga: After starting to work on this I've faced a problem with studio - we don't have any DefinitionRegistryService reference in Studio, only ComponentService is used there.
https://github.com/Talend/tdi-studio-se/blob/ef79d74eb9305e2e630a836ae76fe3b22d685887/main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/utils/ComponentsUtils.java#L92-L102
Should we add something similar to this code for DefinitionRegistry? Getting a reference to DefinitionRegistryService from ComponentService is quite tricky, as we cannot be sure ComponentService will be using DefinitionRegistryService in all implementations.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/897/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/897/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/900/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/900/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/899/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/899/)

PR #781, opened 2017-05-31
feat(TCOMP-507): Unable to get a UI Spec with existing data

charlesnguyen:
https://jira.talendforge.org/browse/TCOMP-507

**What is the new behavior?**
A new end point has been added : 
POST /properties/{definitionName}?formName=MAIN

- [ x] The commit message follows Talend standard

- [ ] Bugfix
- [x] Feature

- [ ] Yes
- [x] No

**Other information**:


services/components-api-service-rest/src/main/java/org/talend/components/service/rest/impl/PropertiesControllerImpl.java
```diff
@@ -190,6 +190,12 @@ public String getProperties(String definitionName, String formName) {
         return new ResponseEntity<>(response, OK);
     }
 
+    public String initializeProperties(String formName, PropertiesDto propertiesContainer) {
+        Properties properties = propertiesHelpers.propertiesFromDto(propertiesContainer);
+        properties.refreshLayout(properties.getForm(formName));
+        return properties == null ? "{}" : jsonSerializationHelper.toJson(formName, properties);
```
sgandon: if the `properties` instance may be null then you have to make sure the code in the line above does not cause an NPE and check for null before calling the *refreshLayout*

services/components-api-service-rest/src/test/java/org/talend/components/service/rest/fullexample/FullExampleFamilyDefinition.java
```diff
@@ -0,0 +1,40 @@
...
+ * Install all of the definitions provided for the FullExample family of components.
+ */
+@AutoService(ComponentInstaller.class)
+@Component(name = Constants.COMPONENT_INSTALLER_PREFIX + FullExampleFamilyDefinition.NAME, provide = ComponentInstaller.class)
+public class FullExampleFamilyDefinition extends AbstractComponentFamilyDefinition implements ComponentInstaller {
+
+    public static final String NAME = "FullExample";
+
+    public FullExampleFamilyDefinition() {
+        super(NAME, new org.talend.components.service.rest.fullexample.FullExampleInputDefinition(), new FullExampleDatastoreDefinition(), new FullExampleDatasetDefinition());
```
sgandon: You don't need the FullExampleInputDefinition for your test so I suggest you remove it and remove the class too (FullExampleInputDefinition.java).

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/454/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/454/)

services/components-api-service-rest/src/test/java/org/talend/components/service/rest/fullexample/FullExampleFamilyDefinition.java
sgandon: And do you need a the Datastore too cause it seems that the test of initialising properly the ui-specs is done on the dataset ?

services/components-api-service-rest/src/main/java/org/talend/components/service/rest/impl/PropertiesControllerImpl.java
charlesnguyen: great point :)

services/components-api-service-rest/src/test/java/org/talend/components/service/rest/fullexample/FullExampleFamilyDefinition.java
charlesnguyen: I test both Datastore  and Dataset.

gonchar-ivan:
As I understand this PR adds new endpoint
POST /properties/{definitionName}
What about documentation for it?
I found we have 
https://github.com/Talend/components/wiki/7.-Components-REST-service#validate-component-properties

but new method is rather _initialize_ method then validate

PR #865, opened 2017-06-28
chore(TCOMP-216): define reader main use cases for tests

akhabali:

**What is the new behavior?**
Define a set of tests that cover the main states of a data source (empty, contain only one element, has multiple element or it unavailable).  those  states should be handled correctly by a Reader .

Every reader Test should implement SourceReaderTest and validate the use cases.

An matchers set is available in the class ReaderMatchers that can be used while testing a Reader (to assert that the reader: canStart(), canAdvance() ...)

An implementation example is available in the package org/talend/components/api/component/runtime/example

In the implementation example you fin a recommended way to code the Reader class by introducing a ReaderService that should wrap all the external api calls to make it easy to test the Reader logic by mocking the ReaderService and emulating the api behavior. 
This make it possible to test the code logic in unit test instead of integration test that depends on external system and that can be unstable. 

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:


build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/677/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/677/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/678/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/678/)

core/components-api/src/test/java/org/talend/components/api/ReaderUseCases.java
```diff
@@ -0,0 +1,262 @@
...
+
+            try {
+                assertTrue(reader.start());
+                assertNotNull(reader.getCurrent());
+                Map<String, Object> returnedValues = reader.getReturnValues();
+                assertNotNull(returnedValues);
+                assertTrue(returnedValues.containsKey(ComponentDefinition.RETURN_TOTAL_RECORD_COUNT));
+                assertEquals(1, returnedValues.get(ComponentDefinition.RETURN_TOTAL_RECORD_COUNT));
+            } catch (Exception e) {
+                fail("expect that the reader starts whit no error in this scenario. Insteade an exception has been thrown"
```
sgandon: typo here *Instead* instead of * Insteade* ;)

core/components-api/src/test/java/org/talend/components/api/ReaderTest.java
```diff
@@ -0,0 +1,74 @@
...
+ *     SourceNotEmpty<IndexedRecord> readerNotEmptyUseCase = new SourceNotEmpty<IndexedRecord>(runtimeContiner, dataSource) {
+ * 
+ *         &#64;Override
+ *         public ComponentProperties initProperties() {
+ *             // initialize the component properties for the test...
+ *             return properties;
+ *         }
+ * 
+ *         &#64;Override
+ *         public Reader<IndexedRecord> mockReaderService() {
```
sgandon: I don't see the reason why you would want to test some mock instance ????? usually you use mock for setup complicated things but testing mocks is useless.

sgandon: If you want to use mocks for testing your how test cases then you can just return a mocked instance during the source.createReader() method.
But no one should use mocks here and therefore please do not have public API expecting mocks.

core/components-api/src/test/java/org/talend/components/api/ReaderUseCases.java
```diff
@@ -0,0 +1,262 @@
...
+
+        /**
+         * Initialize the use case properties
+         */
+        public abstract ComponentProperties initProperties();
+
+        /**
+         * Run the test use case
+         */
+        public abstract void runTest();
```
sgandon: I would rename this starting with an *assert* work because some PMD or checkstyle engine would make test calling this as error cause there is not assert keyword. something like *asserTestRun*

core/components-api/src/test/java/org/talend/components/api/ReaderTest.java
akhabali: The work in this PR is mainly for testing the reader logic.
I have introduced a ReaderService that encapsulate the external API calls (or external system interactions) in my example implementation and this is what should be mocked using this method if necessary.

So the test do cover the Reader logic regarding to a Service Reader mocked responses.

As this service layer doesn't exist in tcomp this method take the reader object to mock it service.
We may simplify this to make it more clear.

What do you think ?

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/681/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/681/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/682/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/682/)

core/components-api/src/test/java/org/talend/components/api/ReaderUseCases.java
```diff
@@ -0,0 +1,262 @@
...
+     * </ul>
+     */
+    public static abstract class SourceNotEmpty<T> extends AbstractReaderUseCase<T> {
+
+        public SourceNotEmpty(RuntimeContainer container, Source source) {
+            super(container, source);
+        }
+
+        @Override
+        public final void asserTestRun() {
```
gonchar-ivan: Oops. 1 't' is missed
I think it should be
`assertTestRun()`

akhabali: agree with you for the package.
we can also have a separated module for test API, but i don't know how this will evolve.

core/components-api/src/test/java/org/talend/components/api/ReaderTest.java
gonchar-ivan: Our problem is following:
1. We have requirement to have unit-test coverage above 60%
2. unit-test coverage doesn't include Integration-tests
Thus, in most cases we can't cover runtime part enough.

As I understand here Anas splits Reader class into 2 parts:
1. code which calls to external API - ReaderService (this part is almost impossible to cover with unit-tests)
2. other code like counting number of lines, throwing exception etc (this part is possible to cover with unit-tests)
Such separation allows to cover a little bit more runtime code.
But still ReaderService implementation for specific component won't be covered in unit-tests. It can be covered only in Integration tests.
Did I understand correctly?

core/components-api/src/test/java/org/talend/components/api/component/runtime/example/DummyReadService.java
```diff
@@ -0,0 +1,40 @@
...
+//
+// ============================================================================
+package org.talend.components.api.component.runtime.example;
+
+import org.apache.avro.generic.IndexedRecord;
+
+/**
+ * A service that encapsulate the interaction with external system / API calls
+ */
+public class DummyReadService {
```
gonchar-ivan: Maybe eventually, we may make it an interface available for all components

core/components-api/src/test/java/org/talend/components/api/ReaderUseCases.java
gonchar-ivan: Yes, eventually it may be separated module

core/components-api/src/test/java/org/talend/components/api/ReaderTest.java
akhabali: it's exactly what i'm trying to achieve here.

the ReaderService should behave as a stateless wrapper of external calls (takes parameters, call api and return result) with no logic or exception handling. 
this make mocking and testing easy but add an additional abstraction layer.
Whit this kind of separation we can simulate api behavior by mocking the ReaderService methods. 

It can have also another advantage as external libraries usage will be encapsulated in a service class then upgrading there versions will impact only that class and not all our readers.

The inconvenient of that is the boilerplate code, but it make testing possible

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/685/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/685/)

core/components-api/src/test/java/org/talend/components/api/ReaderTest.java
gonchar-ivan: I have several thoughts and suggestions about it. I'll try to explain them.

1. Interface with test scenarios methods (ReaderTest). This is good idea. This interface reminds developer about mandatory test scenarios to check. Ideally, in future it may evolve to an abstract class with default implementation for all of these test methods.

2. ReaderService.
ReaderUseCases class introduces a notion of ReaderService. It is also good idea, but it is implicit for the moment (API does not force to implement it). So, developers should follow ReaderService approach to be able to use ReaderUseCases classes. This should be documented very well. Another option is - we may integrate ReaderService to API (I will write about it below). Thus, it will become explicit and standard.

3. ReaderService mock.
Each test scenario requires different ReaderService mock to emulate different conditions (either source is empty or not empty). That's why AbstractReaderUseCase<T> class has abstract mockReaderService() method. This method is supposed to be overridden by developer for each test scenario. initProperties() has similar idea. Developer may override it for every test scenario to emulate different condition.

4. ReaderUseCases.
This is a set of of common test scenarios.
I have an idea about it:
I suggest to create ReaderAssert class (instead of ReaderUseCases) with following static methods (by analogue to Junit Assert class):
```
public static void assertReaderIsEmptry(Reader reader);

public static void assertReaderIsNotEmptry(Reader reader);

public static void assertReaderCanAdvance(Reader reader);

public static void assertReaderCannotAdvance(Reader reader);
```
I would leave Reader creation and ReaderService (note, it may be not ReaderService, but some other approach) mocking up to developer without providing methods for it (for the moment).
Thus, this ReaderAssert can also be used in Integration tests.

5. DummyReader
In current PR classes in org.talend.components.api.component.runtime.example descibes how this test API should be used (also how Reader should be implemented to make it possible to test it with these test API)
This DummyReader shows that we have some common code in Reader class for different components (Exception throwing, counting record number).
This leads to the following idea: we may have CommonReader implementation with several extension points. ReaderService is an example of this extension point. What I mean - developers won't extend Reader interface, but they will extend other classes, which are parts of CommonReader. We need to think about it. If it is good idea, than we can make ReaderService a standard approach and further simplify tests
(hm... thus, we will cover CommonReader once and we won't need tests for it in each comp, but we will need tests for ReaderService and these tests can't be unit-tests)

akhabali: I implemented the ReaderAssert approach described in point 4 that offer more flexibility for the developer to create his Reader. the ReaderAssert test the main use case of a reader (canStart, canAdvance...).

i also redesigned the ReaderTes test interface to SourceReaderTest. this interface provide a suites of test that a reader should handle based on the Data Source State (empty, one element, multiple element, unavailable)

Also moved the class to the suggested package.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/707/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/707/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/705/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/705/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/708/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/708/)

core/components-api/src/test/java/org/talend/components/api/test/runtime/reader/SourceReaderTest.java
```diff
@@ -0,0 +1,63 @@
...
+// %InstallDIR%\features\org.talend.rcp.branding.%PRODUCTNAME%\%PRODUCTNAME%license.txt
+//
+// You should have received a copy of the agreement
+// along with this program; if not, write to Talend SA
+// 9 rue Pages 92150 Suresnes, France
+//
+// ============================================================================
+package org.talend.components.api.test.runtime.reader;
+
+import javax.xml.transform.Source;
```
gonchar-ivan: please remove this import

core/components-api/src/test/java/org/talend/components/api/test/runtime/reader/example/DummyRecord.java
```diff
@@ -0,0 +1,49 @@
...
+// ============================================================================
+package org.talend.components.api.test.runtime.reader.example;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.avro.Schema;
+import org.apache.avro.generic.IndexedRecord;
+
+public class DummyRecord implements IndexedRecord {
```
gonchar-ivan: I suggest to use GenericData.Record() class from Avro library. It has similar implementation

core/components-api/src/test/java/org/talend/components/api/test/runtime/reader/SourceReaderTest.java
akhabali: i don't know way i have this import here 😆 
i removed it

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/714/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/714/)

core/components-api/src/test/java/org/talend/components/api/test/runtime/reader/SourceReaderTest.java
gonchar-ivan: Thanks. 
It was added automatically, because of ```{@link Source}``` in javadocs

core/components-api/src/test/java/org/talend/components/api/test/runtime/reader/ReaderAssert.java
```diff
@@ -0,0 +1,170 @@
...
+     * <ul>
+     * <li>{@link Reader#start()} return <code>true</code> and should not throws any exception</li>
+     * <li>{@link Reader#getCurrent()} return the first read element.</li>
+     * <li>{@link Reader#getReturnValues()} is not null</li>
+     * <li>the value of {@link ComponentDefinition#RETURN_TOTAL_RECORD_COUNT} in the returned map {@link Reader#getReturnValues()}
+     * is equal to 1
+     * </i>
+     * </ul>
+     */
+    public static <T> void canStart(Reader<T> reader) {
```
sgandon: Globally I would suggest you use `assertThat` instead of the basic JUnit asserts, see https://objectpartners.com/2013/09/18/the-benefits-of-using-assertthat-over-other-assert-methods-in-unit-tests/ for some good reasons to do so.

core/components-api/src/test/java/org/talend/components/api/test/runtime/reader/example/DummyReadService.java
```diff
@@ -0,0 +1,44 @@
...
+// along with this program; if not, write to Talend SA
+// 9 rue Pages 92150 Suresnes, France
+//
+// ============================================================================
+package org.talend.components.api.test.runtime.reader.example;
+
+import java.util.List;
+
+/**
+ * This is an example service that wrap the interactions with the external system / API calls.
```
sgandon: you should make it more clear that is in only for testing the this Test framework and not the recommended way for writing Readers.

core/components-api/src/test/java/org/talend/components/api/test/runtime/reader/example/DummyReaderTest.java
```diff
@@ -0,0 +1,204 @@
...
+import org.talend.daikon.properties.ValidationResult;
+
+/**
+ * This an example of {@link SourceReaderTest} implementation that use the {@link ReaderAssert} and benefit from the
+ * Reader/ReaderService design to test the reader logic
+ */
+public class DummyReaderTest implements SourceReaderTest {
+
+    @Mock
+    private DummyReadService readerServiceMock;
```
sgandon: please explain here this is not meant to be used on the real Readers tests but only here to easy the writing of this dummy test.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/863/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/863/)

core/components-api/src/test/java/org/talend/components/api/test/runtime/reader/ReaderMatchers.java
```diff
@@ -0,0 +1,258 @@
...
+                } catch (Throwable e) {
+                    fail("expect that the reader starts whit no error, instead an exception was thrown: "
+                            + ExceptionUtils.getStackTrace(e));
+                }
+                return true;
+            }
+
+            @Override
+            public void describeTo(Description description) {
+                description.appendText("expect that the reader starts whit no error.");
```
sgandon: typo here, not *whit* but *with*

core/components-api/src/test/java/org/talend/components/api/test/runtime/reader/example/DummyReaderTest.java
```diff
@@ -78,8 +86,8 @@ public void testReadSourceEmpty() {
             reader.readerService = readerServiceMock;
 
             // assert
-            ReaderAssert.cannotStart(reader);
-            ReaderAssert.cannotAdvance(reader);
+            assertThat(reader, cannotStart());
```
sgandon: nice ! thanks.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/1036/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/1036/) Sonar results: https://sonar.datapwn.com/overview?id=org.talend.components%3Acomponents%3Aakahabli%2FTCOMP-216-test_usecases

PR #914, opened 2017-07-20
Igonchar/tcomp 603 jdbc caching

gonchar-ivan:

ClassLoader caching is enabled for JDBC components

**What is the new behavior?**

ClassLoader caching is disabled for JDBC components

- [x] Docs have been added / updated (for bug fixes / features) ?

- [x] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)

- [ ] Yes
- [x] No

**Other information**:
Related to https://jira.talendforge.org/browse/TCOMP-603


build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/850/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/850/)

gonchar-ivan:
Depends on https://github.com/Talend/daikon/pull/256

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/854/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/854/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/868/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/868/)

francisu:
Why are we using Mocks? We should be testing this with actual databases. Even the basic tests should use something like HSQL which is simple. We should have tests for all drivers that we support with the actual databases.

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
```diff
@@ -17,57 +17,81 @@
...
+        }
         this.props = props;
+        if (props.getRuntimeSetting().getDriverClass() == null) {
+            throw new NullPointerException("props must provide not null driver class");
+        }
+        this.driverClassName = props.getRuntimeSetting().getDriverClass();
     }
 
-    public JarRuntimeInfo cloneWithNewJarUrlString(String newJarUrlString) {
+    public JdbcRuntimeInfo cloneWithNewJarUrlString(String newJarUrlString) {
```
wwang-talend: it will break the code in talend studio as we call this method there and no compiler time dependency, please see : 
org.talend.designer.core.generic.model.Component

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
```diff
@@ -79,4 +103,33 @@ String removeQuote(String content) {
...
+    /**
+     * Checks whether this instance equals to <code>obj</code>.
+     * {@link #driverClassName} also is taken into account during comparison
+     */
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (!JarRuntimeInfo.class.isAssignableFrom(obj.getClass())) {
```
wwang-talend: what's the difference with "instanceof"? in my opinion, "instanceof" is more easy to understand, not important, you can ignore this comment.

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
wwang-talend: seems i am wrong, please ignore the comment above.

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
ypiel-talend: Some differences explained here : https://stackoverflow.com/questions/496928/what-is-the-difference-between-instanceof-and-class-isassignablefrom

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
```diff
@@ -79,4 +103,33 @@ String removeQuote(String content) {
...
+            return false;
+        }
+        JdbcRuntimeInfo other = (JdbcRuntimeInfo) obj;
```
ypiel-talend: Indeed, why the above test checks the super class JarRuntimeInfo ? And here obj is casted in current class JdbcRuntimeInfo ?
If the aim is to cast in JdbcRuntimeInfo the above test should be 'if (!(obj instanceof JdbcRuntimeInfo ))' isn'it ?

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
```diff
@@ -79,4 +103,33 @@ String removeQuote(String content) {
...
+        return super.equals(obj) && this.driverClassName.equals(other.driverClassName);
+    }
+
+    @Override
+    public String toString() {
+        return super.toString() + ", " + "JdbcRuntimeInfo: {" + "driverClassName:" + driverClassName + "}";
```
ypiel-talend: Why create so much strings ? ", " + "JdbcRuntimeInfo: {" + "driverClassName:"

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
gonchar-ivan: yes. You are right!
I just copied and pasted code from super class and forgot to change JarRuntimeInfo to JdbcRuntimeInfo.
I will change that check to
`obj instanceof JdbcRuntimeInfo`

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
gonchar-ivan: yes you right
I will change to a single string.

However, in any case this is not a big problem (it was a problem some time ago).
Actually, compiler do 2 things, which optimizes code:
1. It concatenates these strings into one string (LDC ", JdbcRuntimeInfo: {driverClassName:")
2. It uses StringBuilder to create this big string
I have some bytecode plugin in my Eclipse, which shows bytecode representation of the method:

```
 LINENUMBER 132 L0
    NEW java/lang/StringBuilder
    DUP
    INVOKESPECIAL java/lang/StringBuilder.<init> ()V
    ALOAD 0
    INVOKESPECIAL org/talend/components/api/component/runtime/JarRuntimeInfo.toString ()Ljava/lang/String;
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    LDC ", JdbcRuntimeInfo: {driverClassName:"
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    ALOAD 0
    GETFIELD org/talend/components/jdbc/JdbcRuntimeInfo.driverClassName : Ljava/lang/String;
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    LDC "}"
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    INVOKEVIRTUAL java/lang/StringBuilder.toString ()Ljava/lang/String;
    ARETURN
```

So, in some cases we may split a big string to smaller strings in source code, if it improves readability. (I agree there is no sense to do it in this case.)

components/components-jdbc/components-jdbc-definition/src/test/java/org/talend/components/jdbc/JdbcRuntimeInfoTest.java
```diff
@@ -1,21 +1,156 @@
...
+        RuntimeSettingProvider runtimeSettingProvider1 = createPropsWithDriverName("org.talend.Driver1");
+        JdbcRuntimeInfo jarRuntimeInfo1 = new JdbcRuntimeInfo(runtimeSettingProvider1, "anyRuntimeClassName");
+        int hashCode1 = jarRuntimeInfo1.hashCode();
+
+        RuntimeSettingProvider runtimeSettingProvider2 = createPropsWithDriverName("org.talend.Driver1");
+        JdbcRuntimeInfo jarRuntimeInfo2 = new JdbcRuntimeInfo(runtimeSettingProvider2, "anyRuntimeClassName");
+        int hashCode2 = jarRuntimeInfo2.hashCode();
+
+        assertEquals(269161086, hashCode1);
+        assertEquals(269161086, hashCode2);
```
ypiel-talend: Should those hashcodes be hardcoded ? They depend on HashCodeBuilder.toHashCode() implementation.

components/components-jdbc/components-jdbc-definition/src/test/java/org/talend/components/jdbc/JdbcRuntimeInfoTest.java
```diff
@@ -1,21 +1,156 @@
...
+        RuntimeSettingProvider runtimeSettingProvider1 = createPropsWithDriverName("org.talend.Driver1");
+        JdbcRuntimeInfo jarRuntimeInfo1 = new JdbcRuntimeInfo(runtimeSettingProvider1, "anyRuntimeClassName");
+        int hashCode1 = jarRuntimeInfo1.hashCode();
+
+        RuntimeSettingProvider runtimeSettingProvider2 = createPropsWithDriverName("org.talend.Driver2");
+        JdbcRuntimeInfo jarRuntimeInfo2 = new JdbcRuntimeInfo(runtimeSettingProvider2, "anyRuntimeClassName");
+        int hashCode2 = jarRuntimeInfo2.hashCode();
+
+        assertEquals(269161086, hashCode1);
+        assertEquals(269161087, hashCode2);
```
ypiel-talend: Should those hashcodes be hardcoded ? They depend on HashCodeBuilder.toHashCode() implementation.

components/components-jdbc/components-jdbc-definition/src/test/java/org/talend/components/jdbc/JdbcRuntimeInfoTest.java
```diff
@@ -1,21 +1,156 @@
...
+    @Test
+    public void testEquals() {
+        RuntimeSettingProvider runtimeSettingProvider1 = createPropsWithDriverName("org.talend.Driver1");
+        JdbcRuntimeInfo jarRuntimeInfo1 = new JdbcRuntimeInfo(runtimeSettingProvider1, "anyRuntimeClassName");
+
+        RuntimeSettingProvider runtimeSettingProvider2 = createPropsWithDriverName("org.talend.Driver1");
+        JdbcRuntimeInfo jarRuntimeInfo2 = new JdbcRuntimeInfo(runtimeSettingProvider2, "anyRuntimeClassName");
+
+        assertTrue(jarRuntimeInfo1.equals(jarRuntimeInfo2));
+        assertTrue(jarRuntimeInfo2.equals(jarRuntimeInfo1));
```
ypiel-talend: If one of those assert is true the second 'must' be true too since equals must be symmetric.

Maybe other rules of the equals contract should be tested ?https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)

gonchar-ivan: Yes, I checked symmetric rule.
Yes, I will add tests for other rules from equals contract

components/components-jdbc/components-jdbc-definition/src/test/java/org/talend/components/jdbc/JdbcRuntimeInfoTest.java
```diff
@@ -1,21 +1,156 @@
...
+
+    /**
+     * Checks {@link JdbcRuntimeInfo#cloneWithNewJarUrlString(String)} creates new instance of {@link JdbcRuntimeInfo}
+     */
+    @Test
+    public void testCloneWithNewJarUrlString() {
+        RuntimeSettingProvider properties = createPropsWithDriverName("org.talend.Driver1");
+        JdbcRuntimeInfo jdbcRuntimeInfo = new JdbcRuntimeInfo(properties, "anyRuntimeClassName");
+        JdbcRuntimeInfo clone = jdbcRuntimeInfo.cloneWithNewJarUrlString("mvn: new/url");
+        assertTrue(jdbcRuntimeInfo != clone);
```
ypiel-talend: Only test it is a new instance. Maybe it should test other rules of clone contract ? Maybe assertTrue(jdbcRuntimeInfo.equals(clone)) ?
https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#clone()

gonchar-ivan: mmm... this is not real clone() method.
It creates a new instance with new value. So, equals won't return true.

ypiel-talend: Ho yes, it's not clone()... equals will be false. The values that don't change shouldn't be tested ?

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
ypiel-talend: ok, I knew about compiler String optimization but this concatenation sounds strange for me :) great eclipse plugin, thanks.

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
gonchar-ivan: done

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
gonchar-ivan: done

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
gonchar-ivan: done

components/components-jdbc/components-jdbc-definition/src/test/java/org/talend/components/jdbc/JdbcRuntimeInfoTest.java
gonchar-ivan: done

components/components-jdbc/components-jdbc-definition/src/test/java/org/talend/components/jdbc/JdbcRuntimeInfoTest.java
gonchar-ivan: done

components/components-jdbc/components-jdbc-definition/src/test/java/org/talend/components/jdbc/JdbcRuntimeInfoTest.java
gonchar-ivan: done (reflexive, transitive, null value)
symmetric already exist.
I think there is no sense for consistent (it will look strange)

components/components-jdbc/components-jdbc-definition/src/test/java/org/talend/components/jdbc/JdbcRuntimeInfoTest.java
gonchar-ivan: added more asserts to compare state of original and "cloned" instances

components/components-jdbc/components-jdbc-definition/src/main/java/org/talend/components/jdbc/JdbcRuntimeInfo.java
gonchar-ivan: Usually, change return type from super to child class is safe (I don't know any case, when it may break code).
It shouldn't break code here. Anyway thanks for comment)))

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/874/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/874/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/873/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/873/)

PR #931, opened 2017-08-18
Mbasiuk/tdi 39295 snowflake row

mbasiuk-talend:
No tSnowflakeRow component.

**What is the new behavior?**
Create and implement tSnowflakeRow component.

- [x] The new code does not introduce new technical issues (sonar / eslint)

- [ ] Bugfix
- [x] Feature

- [ ] Yes
- [x] No

**Other information**:


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/945/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/945/)

components/components-snowflake/src/main/java/org/talend/components/snowflake/SnowflakePreparedStatementTable.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.commons.lang3.reflect.TypeLiteral;
+import org.talend.components.api.properties.ComponentPropertiesImpl;
+import org.talend.daikon.properties.presentation.Form;
+import org.talend.daikon.properties.presentation.Widget;
+import org.talend.daikon.properties.property.Property;
+
+public class SnowflakePreparedStatementTable extends ComponentPropertiesImpl {
```
dmytro-sylaiev: It would be great if you specify in class name that it is properties

components/components-snowflake/src/main/java/org/talend/components/snowflake/SnowflakePreparedStatementTable.java
```diff
@@ -0,0 +1,71 @@
...
+        mainForm.addColumn(values);
+    }
+
+    @Override
+    public void setupProperties() {
+        super.setupProperties();
+        types.setPossibleValues(Arrays.asList(Type.values()));
+    }
+
+    public enum Type {
```
dmytro-sylaiev: It seems to me that this types (and whole properties) will be the same or at least similar for all db row' components
You can in advance move it in the some parent class

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private Result result;
+
+    private transient JDBCResultSetIndexedRecordConverter factory = new SnowflakeResultSetIndexedRecordConverter();
+
+    private final Boolean usePreparedStatement;
+
+    private PreparedStatement pstmt;
+
+    private Statement statement;
```
dmytro-sylaiev: interface PreparedStatement extends Statement, it seems you don't use both variables simultaneously. So you can remain only one of them

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Override
+    public Instant getCurrentTimestamp() {
+        return Instant.now();
+    }
+
+    @Override
+    public void close() throws IOException {
+        try {
+            if (rs != null) {
```
dmytro-sylaiev: According to javadoc, 

> <B>Note:</B>When a <code>Statement</code> object is
>      * closed, its current <code>ResultSet</code> object, if one exists, is
>      * also closed.

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowStandalone.java
```diff
@@ -0,0 +1,97 @@
...
+            }
+        } catch (SQLException e) {
+            if (dieOnError) {
+                throw new ComponentException(e);
+            }
+            LOGGER.error(I18N_MESSAGES.getMessage("error.queryExecution"), e);
+        } catch (IOException ioe) {
+            if (dieOnError) {
+                throw new ComponentException(ioe);
+            }
```
dmytro-sylaiev: Just proposal: you can move this if clause in some `private void throwComponentException(Exception e)` method

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/utils/SnowflakePreparedStatementUtils.java
```diff
@@ -0,0 +1,102 @@
...
+                pstmt.setShort(index, (short) value);
+                break;
+            case String:
+                pstmt.setString(index, (String) value);
+                break;
+            case Time:
+                pstmt.setTime(index, (Time) value);
+                break;
+            case Null:
+                pstmt.setNull(index, (int) value);
```
dmytro-sylaiev: It is not obviously for me why do you set value as sqlType
Could you please elaborate on it?

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
```diff
@@ -0,0 +1,190 @@
...
+
+    protected transient PropertyPathConnector REJECT_CONNECTOR = new PropertyPathConnector(Connector.REJECT_NAME, "schemaReject");
+
+    public SchemaProperties schemaFlow = new SchemaProperties("schemaFlow");
+
+    public SchemaProperties schemaReject = new SchemaProperties("schemaReject");
+
+    public final PresentationItem guessQuery = new PresentationItem("guessQuery");
+
+    public Property<String> query = PropertyFactory.newString("query", "\"select id, name from employee\"");
```
dmytro-sylaiev: Note for future, when using existing connection with tableName set, or create component from wizard with tablename, we can already set tablename in the query inside `employee`

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/953/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/953/)

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
mbasiuk-talend: This is a template. User can perform guessQuery for creating a valid query.

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
mbasiuk-talend: "It is generally good practice to release resources as soon as you are finished with them to avoid tying up database resources."
Connection may be closed later (while using existing connection). That's why it's not a good choice.

components/components-snowflake/src/main/java/org/talend/components/snowflake/SnowflakePreparedStatementTable.java
mbasiuk-talend: Agree.

components/components-snowflake/src/main/java/org/talend/components/snowflake/SnowflakePreparedStatementTable.java
mbasiuk-talend: It's a good idea, but not a part of this task. I will mark with TODO comment.

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
mbasiuk-talend: It's a good way to close them in defined order ResultSet/Statement/Connection.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/956/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/956/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/965/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/965/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/988/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/988/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/989/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/989/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/992/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/992/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/999/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/999/)

components/components-snowflake/src/main/java/org/talend/components/snowflake/SnowflakePreparedStatementTableProperties.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * This class is responsible for inserting proper values into prepared statement used in {@link TSnowflakeRowDefinition} component<br/>
+ * It contains: <b>Indexes</b>, <b>Types</b>, <b>Values</b>.<br/>
+ * Field <b>Indexes</b> is responsible for specifying a correct index in prepared statement.<br/>
+ * <b>Type</b> is responsible for database type used for inserting value in prepared statement.<br/>
+ * <b>Values</b> actual values to be inserted into prepared statement.
+ *
+ */
+public class SnowflakePreparedStatementTableProperties extends ComponentPropertiesImpl {
```
gonchar-ivan: This Properties can be common for several Components (At least for Snowflake and JDBC)
So, I suggest:
1. Rename it to PreparedStatementProperties
2. Eventually move it to components-common module
I found the same class and implementation in JDBC component (PreparedStatementTable class in components-jdbc-definition)

components/components-snowflake/src/main/java/org/talend/components/snowflake/SnowflakePreparedStatementTable.java
gonchar-ivan: TODOs doesn't work. Will you be able to remind all your TODOs after some time?
Better to create Jira issue for this task. Also better to try to do it in scope of current task. Otherwise, it won't be done at all. We will never have time to work on such issues.

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
```diff
@@ -0,0 +1,228 @@
...
+
+
+    public TSnowflakeRowProperties(String name) {
+        super(name);
+    }
+
+    @Override
+    public void setupProperties() {
+        super.setupProperties();
+    }
```
gonchar-ivan: You may remove remove this method. Nothing will change

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
```diff
@@ -0,0 +1,228 @@
...
+ *
+ *
+ */
+public class TSnowflakeRowProperties extends SnowflakeConnectionTableProperties {
+
+    private static final long serialVersionUID = -1694372250903352577L;
+
+    protected transient PropertyPathConnector FLOW_CONNECTOR = new PropertyPathConnector(Connector.MAIN_NAME, "schemaFlow");
+
+    protected transient PropertyPathConnector REJECT_CONNECTOR = new PropertyPathConnector(Connector.REJECT_NAME, "schemaReject");
```
gonchar-ivan: seems, there is sense to make connectors static and final.
What do you think about it? (We need to consider serialization/deserialization here and migration, whether it affects it)
Why these properties are `protected`? Do you expect somebody will extend this class?

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
```diff
@@ -0,0 +1,228 @@
...
+
+    public final PresentationItem guessQuery = new PresentationItem("guessQuery");
+
+    public Property<String> query = PropertyFactory.newString("query", "\"select id, name from employee\"");
+
+    public Property<Boolean> dieOnError = PropertyFactory.newBoolean("dieOnError").setRequired();
+
+    // advanced
+
+    public Property<Boolean> propagateQueryResultSet = PropertyFactory.newBoolean("propagateQueryResultSet").setRequired();
```
gonchar-ivan: I'm not sure this property is correct for new components.
It was possible for old components to propagate instance of ResultSet to next component.
For new components it is not fully correct to do it, because Avro doesn't know what to do with ResultSet (however it may be possible by some workaround, but only for Studio)

gonchar-ivan: How does "setRequired" affects on property?

gonchar-ivan:
Please, check component icon. As I see most *Row components have icon with lightning. 
Is there a Jira issue for UX team to draw a new icon for tJDBCRow component?

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
mbasiuk-talend: After your comment, I think to remove this property. I totally agree with you that we should use Avro types for new components.

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
```diff
@@ -0,0 +1,228 @@
...
+
+    private static final long serialVersionUID = -1694372250903352577L;
+
+    protected transient PropertyPathConnector FLOW_CONNECTOR = new PropertyPathConnector(Connector.MAIN_NAME, "schemaFlow");
+
+    protected transient PropertyPathConnector REJECT_CONNECTOR = new PropertyPathConnector(Connector.REJECT_NAME, "schemaReject");
+
+    public SchemaProperties schemaFlow = new SchemaProperties("schemaFlow");
+
+    public SchemaProperties schemaReject = new SchemaProperties("schemaReject");
```
gonchar-ivan: Maybe schemaFlow and schemaReject can be stored as Schema class instances?
Or is it required for PropertyPathConnectors?

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
```diff
@@ -0,0 +1,228 @@
...
+                schema.isError());
+        List<Schema.Field> copyFieldList = new ArrayList<>();
+        for (Schema.Field se : schema.getFields()) {
+            Schema.Field field = new Schema.Field(se.name(), se.schema(), se.doc(), se.defaultVal(), se.order());
+            field.getObjectProps().putAll(se.getObjectProps());
+            for (Map.Entry<String, Object> entry : se.getObjectProps().entrySet()) {
+                field.addProp(entry.getKey(), entry.getValue());
+            }
+            copyFieldList.add(field);
+        }
```
gonchar-ivan: I often see similar code, which creates new schema from old one with several additional fields.
So, I've created following Jiras: 
https://jira.talendforge.org/browse/TDKN-174
https://jira.talendforge.org/browse/TDKN-96

but I haven't implemented them yet (

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
```diff
@@ -0,0 +1,228 @@
...
+        rejectField.addProp(SchemaConstants.TALEND_IS_LOCKED, "false");
+        rejectField.addProp(SchemaConstants.TALEND_FIELD_GENERATED, "true");
+        rejectField.addProp(SchemaConstants.TALEND_COLUMN_DB_LENGTH, "255");
+        additionalRejectFields.add(rejectField);
+
+        rejectField = new Schema.Field("errorMessage", Schema.create(Schema.Type.STRING), null, (Object) null);
+        rejectField.addProp(SchemaConstants.TALEND_IS_LOCKED, "false");
+        rejectField.addProp(SchemaConstants.TALEND_FIELD_GENERATED, "true");
+        rejectField.addProp(SchemaConstants.TALEND_COLUMN_DB_LENGTH, "255");
+        additionalRejectFields.add(rejectField);
```
gonchar-ivan: maybe there is a sense to extract method:
```
private Schema.Field createRejectField(String fieldName) {
   Schema.Field rejectField = new Schema.Field(fieldName, Schema.create(Schema.Type.STRING), null, (Object) null);
   ...
}
```

Then, your code will look like:
```
final List<Schema.Field> additionalRejectFields = new ArrayList<Schema.Field>();
additionalRejectFields.add(createRejectField("errorCode"));
additionalRejectFields.add(createRejectField("errorMessage"));
```

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
```diff
@@ -0,0 +1,138 @@
...
+ * It can be used to retrieve data using simple query or prepared statement.
+ *
+ */
+public class SnowflakeRowReader implements Reader<IndexedRecord> {
+
+    private final RuntimeContainer container;
+
+    private final SnowflakeRowSource source;
+
+    private final TSnowflakeRowProperties rowProperties;
```
gonchar-ivan: properties are already stored in source.
storing them also in this class seems redundant for me.

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Override
+    public boolean start() throws IOException {
+        connection = source.connect(container);
+        result = new Result();
+        Schema schema = rowProperties.schemaFlow.schema.getValue();
+        if (AvroUtils.isIncludeAllFields(schema)) {
+            schema = source.getEndpointSchema(container, rowProperties.table.tableName.getValue());
+        }
```
gonchar-ivan: I suggest to add methods for retrieving property value in Source class:
```
Schema getRuntimeSchema(RuntimeContainer container) {
    Schema schema = rowProperties.schemaFlow.schema.getValue();
        if (AvroUtils.isIncludeAllFields(schema)) {
            schema = getEndpointSchema(container, rowProperties.table.tableName.getValue());
       }
}
```

```
String getPreparedStatementQuery() {
    return rowProperties.query.getValue();
}
```
etc

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
```diff
@@ -0,0 +1,138 @@
...
+        } catch (SQLException e) {
+            throw new IOException(e);
+        }
+        return advance();
+    }
+
+    @Override
+    public boolean advance() throws IOException {
+        try {
+            return rs.next();
```
gonchar-ivan: It is recommended that this method also converts data to IndexedRecord,
because getCurrent() could be called several times (according javadocs). We need to avoid data conversion for each call.

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
```diff
@@ -0,0 +1,138 @@
...
+        } catch (SQLException e) {
+            throw new IOException(e);
+        }
+    }
+
+    @Override
+    public IndexedRecord getCurrent() {
+        result.totalCount++;
```
gonchar-ivan: This counter should be incremented in advance() method
getCurrent() could be called multiple times between calling advance()

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private final SnowflakeRowSource source;
+
+    private final TSnowflakeRowProperties rowProperties;
+
+    private Connection connection;
+
+    private Result result;
+
+    private transient JDBCResultSetIndexedRecordConverter factory = new SnowflakeResultSetIndexedRecordConverter();
```
gonchar-ivan: please, rename this field to `converter` to not confuse it with Factory pattern

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowSink.java
```diff
@@ -0,0 +1,50 @@
...
+    private static final long serialVersionUID = -6866612725653393551L;
+
+    private TSnowflakeRowProperties rowProperties;
+
+    @Override
+    public ValidationResult initialize(RuntimeContainer container, ComponentProperties properties) {
+        super.initialize(container, properties);
+        this.rowProperties = (TSnowflakeRowProperties) properties;
+        return ValidationResult.OK;
+    }
```
gonchar-ivan: Maybe there is a sense to extract SnowflakeRowSourceOrSink superclass with initialize() and getRowProperties() methods?
What do you think?

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowStandalone.java
gonchar-ivan: Also you may include error code as parameter:
```
private void throwComponentException(Exception e, String errorCode)
```

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/utils/SnowflakePreparedStatementUtils.java
```diff
@@ -0,0 +1,102 @@
...
+     * @throws SQLException
+     */
+    public static void fillPreparedStatement(PreparedStatement pstmt, SnowflakePreparedStatementTableProperties preparedStatementTable)
+            throws SQLException {
+        List<Integer> indexes = preparedStatementTable.indexes.getValue();
+        List<String> types = preparedStatementTable.types.getValue();
+        List<Object> values = preparedStatementTable.values.getValue();
+        for (int i = 0; i < indexes.size(); i++) {
+            Integer index = indexes.get(i);
+            SnowflakePreparedStatementTableProperties.Type type = SnowflakePreparedStatementTableProperties.Type.valueOf(types.get(i));
```
gonchar-ivan: Theoretically, you can avoid big switch-case by adding following method to Type enum:
```
void setPreparedStatementParameter(int index, PreparedStatement statement, Object value);
```
BigDecimal type will have following implementation:
```
void setPreparedStatementParameter(int index, PreparedStatement statement, Object value) {
    statement.setBigDecimal(index, (BigDecimal) value);
}
```

but I'm not sure whether it is good solution for enum (we are implementing some logic in enum instance)

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,250 @@
...
+
+    private Statement statement;
+
+    private ResultSet rs;
+
+    private IndexedRecordConverter<Object, ? extends IndexedRecord> factory;
+
+    private Result result;
+
+    public SnowflakeRowWriter(RuntimeContainer adaptor, WriteOperation<Result> writeOperation) {
```
gonchar-ivan: I think SnowflakeRowWriter can't work with any WriteOperation, so incoming parameter type should be changed to SnowflakeRowWriteOperation.

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,250 @@
...
+        this.container = adaptor;
+        this.writeOperation = writeOperation;
+        this.sink = (SnowflakeRowSink) writeOperation.getSink();
```
gonchar-ivan: SnowflakeRowWriteOperation.getSink() method return type may be changed to SnowflakeRowSink, thus cast won't be required

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,250 @@
...
+
+    private final TSnowflakeRowProperties rowProperties;
+
+    private Connection connection;
+
+    private final Boolean usePreparedStatement;
+
+    private final Boolean propagateQueryResultSet;
+
+    private final Boolean dieOnError;
```
gonchar-ivan: above boolean fields duplicates values stored in TSnowflakeRowProperties  instance. 
I suggest you to add convenient methods in TSnowflakeRowProperties class to retrieve these methods:
```
boolean usePreparedStatement();
boolean propagateQueryResultSet();
boolean dieOnError();
```

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,250 @@
...
+
+    private final int commitEvery;
+
+    private int commitCount;
```
gonchar-ivan: same thing here. Please add convenient methods to Properties class and call these methods whenever you need values

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
gonchar-ivan: Please, rename parameter adaptor -> container
(Actually, I would even prefer `context` (RuntimeContext class), but we class name is RuntimeContainer, so we need to use container name)

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,250 @@
...
+    public SnowflakeRowWriter(RuntimeContainer adaptor, WriteOperation<Result> writeOperation) {
+        this.container = adaptor;
+        this.writeOperation = writeOperation;
+        this.sink = (SnowflakeRowSink) writeOperation.getSink();
+        this.rowProperties = sink.getRowProperties();
+        this.usePreparedStatement = rowProperties.usePreparedStatement.getValue();
+        this.propagateQueryResultSet = rowProperties.propagateQueryResultSet.getValue();
+        this.dieOnError = rowProperties.dieOnError.getValue();
+        this.commitEvery = rowProperties.commitCount.getValue();
+    }
```
gonchar-ivan: Above 5 values can be retrieved from Properties instance at any time

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,250 @@
...
+        successfulWrites = new ArrayList<>();
+        rejectedWrites = new ArrayList<>();
+
+        connection = sink.connect(container);
+        try {
+            if (usePreparedStatement) {
+                statement = connection.prepareStatement(rowProperties.query.getValue());
+            } else {
+                statement = connection.createStatement();
+            }
```
gonchar-ivan: if Sink class will be responsible for keeping connection instance, then above code can also be moved to Sink.getPreparedStatement() method

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,250 @@
...
+
+    @SuppressWarnings({ "unchecked", "deprecation" })
+    @Override
+    public void write(Object object) throws IOException {
+        result.totalCount++;
+        commitCount++;
+        if (null == factory) {
+            factory = (IndexedRecordConverter<Object, ? extends IndexedRecord>) SnowflakeAvroRegistry.get()
+                    .createIndexedRecordConverter(object.getClass());
+        }
```
gonchar-ivan: I suggest to rename factory to converter and extract this code to method:
getConverter()

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,250 @@
...
+
+        IndexedRecord input = factory.convertToAvro(object);
```
gonchar-ivan: incoming object is always instance of IndexedRecord. You don't need to use some converter to convert it. Just cast it to IndexedRecord

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,250 @@
...
+
+        try {
+            if (usePreparedStatement) {
```
gonchar-ivan: This if-else block is used twice in this class.
This makes me to think this class can be divided into 2 subclasses:
StatementWriter and PreparedStatementWriter
WriteOperation may decide which object to create according Properties value.

gonchar-ivan: or there might be special Strategy class for handling Statements.
WriteOperation may pass correct Strategy to Writer constructor.
This options seems better for me as there is a lot of common code in open() and write() methods

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,250 @@
...
+            if (rowProperties.connection.getReferencedComponentId() == null && commitCount == commitEvery) {
+                connection.commit();
+                commitCount = 0;
+            }
+        } catch (SQLException e) {
+            if (dieOnError) {
+                throw new IOException(e);
+            }
+            LOGGER.error(I18N_MESSAGES.getMessage("error.performCommit"), e);
+        }
```
gonchar-ivan: I recommend to extract this code to separate method: commitIfRequired()

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/1013/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/1013/) Sonar results: [https://sonar.datapwn.com/overview?id=org.talend.components%3Acomponents%3Ambasiuk%2FTDI-39295_snowflake_row]

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/utils/SnowflakeRuntimeHelper.java
```diff
@@ -0,0 +1,107 @@
...
+
+
+    public static void closeConnection(RuntimeContainer container, Connection conn, SnowflakeConnectionProperties connProps) throws SQLException {
+        String refComponentId = connProps.getReferencedComponentId();
+        if ((refComponentId == null || container == null) && (conn != null && !conn.isClosed())) {
+            conn.close();
+        }
+    }
+
+    public static Connection getConnection(SnowflakeConnectionProperties connProps, Driver driver) throws Exception{
```
gonchar-ivan: I would rename it to `createConnection`

gonchar-ivan: Why it declares throws Exception? Seems, SQLException is enough

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/utils/SnowflakeRuntimeHelper.java
```diff
@@ -0,0 +1,107 @@
...
+            // Design time
+            connProps = connProps.getReferencedConnectionProperties();
+            // FIXME This should not happen - but does as of now
+            if (connProps == null) {
+                throw new IOException(I18N_MESSAGES.getMessage("error.refComponentWithoutProperties", refComponentId));
+            }
+        }
+
+        try {
+            Driver driver = (Driver) Class.forName(SnowflakeConstants.SNOWFLAKE_DRIVER).newInstance();
```
gonchar-ivan: Is it really required to call Class.forName?

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/utils/SnowflakeRuntimeHelper.java
```diff
@@ -0,0 +1,107 @@
...
+
+
+    public static final String KEY_CONNECTION = "Connection";
+
+    public static final String KEY_CONNECTION_PROPERTIES = "ConnectionProperties";
+
+    private SnowflakeRuntimeHelper() {
+    }
+
+    public static Connection connect(RuntimeContainer container, SnowflakeConnectionProperties connProps) throws IOException {
```
gonchar-ivan: I suggest to extract 2 methods out of this code:
getExistingConnection()
and
createNewConnection() // if there is no existing

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    @Override
+    public void afterMainSchema() {
+        updateOutputSchema();
+        updateRejectSchema();
+    }
+
+    private String generateSqlQueryFromSchema(String rawTableName, Schema schema) {
+        String tableName = StringUtils.strip(rawTableName, "\"");
```
gonchar-ivan: it is already stripped in validateGuessQuery() method

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
```diff
@@ -0,0 +1,203 @@
...
+        schemaReject.schema.setValue(newSchema);
+    }
+
+    @Override
+    public void afterMainSchema() {
+        updateOutputSchema();
+        updateRejectSchema();
+    }
+
+    private String generateSqlQueryFromSchema(String rawTableName, Schema schema) {
```
gonchar-ivan: It is clear from incoming parameter (Schema schema) that this method generates SQL query out of Schema instance.
so, "FromSchema" can be omitted
`generateSqlQuery` seems to be better name

components/components-snowflake/src/test/java/org/talend/components/snowflake/runtime/SnowflakeRowReaderTest.java
```diff
@@ -0,0 +1,184 @@
...
+        Mockito.when(statement.executeQuery(Mockito.eq(query))).thenReturn(rs);
+        Mockito.when(rs.next()).thenReturn(true, false);
+
+        try {
+            do {
+                if (reader.start()) {
+                    IndexedRecord indexedRecord = reader.getCurrent();
+                    Assert.assertEquals(FIELD_VALUE, indexedRecord.get(indexedRecord.getSchema().getField("name").pos()));
+                }
+            } while (reader.advance());
```
gonchar-ivan: seems, this do-while loop is wrong. It calls reader.start() method each iteration.
One of unit-tests purpose is documentation. Use-cases shows class usage. But this use-case shows wrong usage.

components/components-snowflake/src/test/java/org/talend/components/snowflake/runtime/SnowflakeRowReaderTest.java
```diff
@@ -0,0 +1,184 @@
...
+        Mockito.when(connection.createStatement()).thenReturn(statement);
+        rs = Mockito.mock(ResultSet.class);
+        ResultSetMetaData rsMetadata = Mockito.mock(ResultSetMetaData.class);
+        Mockito.when(rs.getMetaData()).thenReturn(rsMetadata);
+        Mockito.when(rsMetadata.getColumnCount()).thenReturn(1);
+        Mockito.when(rs.getString(1)).thenReturn(FIELD_VALUE);
+        Mockito.when(statement.executeQuery(Mockito.eq(query))).thenReturn(rs);
+        Mockito.when(rs.next()).thenReturn(true, false);
+
+        try {
```
gonchar-ivan: try-catch seems redundant here. Just declare that test case may throw IOException

components/components-snowflake/src/test/java/org/talend/components/snowflake/runtime/SnowflakeRowReaderTest.java
```diff
@@ -0,0 +1,184 @@
...
+        Mockito.when(rs.next()).thenReturn(true, true, true, false);
+
+        ResultSetMetaData rsMetadata = Mockito.mock(ResultSetMetaData.class);
+        Mockito.when(rs.getMetaData()).thenReturn(rsMetadata);
+        Mockito.when(rsMetadata.getColumnCount()).thenReturn(remoteSchema.getFields().size());
+        Mockito.when(rs.getInt(1)).thenReturn(1, 2, 3);
+        Mockito.when(rs.getObject(1)).thenReturn(1, 2, 3);
+        Mockito.when(rs.getString(2)).thenReturn(FIELD_VALUE);
+        Mockito.when(rs.getInt(3)).thenReturn(20);
+        Mockito.when(rs.getObject(3)).thenReturn(20);
```
gonchar-ivan: This is an example when it is easier to implement integration test, than to create all these mocks.

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/1018/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/1018/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/1043/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/1043/) Sonar results: https://sonar.datapwn.com/overview?id=org.talend.components%3Acomponents%3Ambasiuk%2FTDI-39295_snowflake_row

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1051/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1051/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1059/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1059/)

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowProperties.java
mbasiuk-talend: Need to check. If it is ok, gonna change it too private

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/utils/SnowflakeRuntimeHelper.java
mbasiuk-talend: Didn't check it, just moved from original location to this helper class.

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
mbasiuk-talend: Ok, moved.

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowReader.java
mbasiuk-talend: Aggree.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1061/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1061/)

components/components-snowflake/src/main/java/org/talend/components/snowflake/SnowflakeConnectionTableProperties.java
```diff
@@ -56,6 +63,10 @@ public void setupLayout() {
         advancedForm.addRow(connection.getForm(Form.ADVANCED));
     }
 
+    public void afterMainSchema() {
+        //Implement in subclasses.
+    };
```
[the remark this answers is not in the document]
mbasiuk-talend: Good catch!

components/components-snowflake/src/main/java/org/talend/components/snowflake/SnowflakePreparedStatementTableProperties.java
mbasiuk-talend: Need to discuss it.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1065/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1065/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1066/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1066/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1068/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1068/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1069/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1069/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1073/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1073/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1074/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1074/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1075/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1075/)

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowStandalone.java
```diff
@@ -0,0 +1,97 @@
...
+
+        Connection connection = null;
+        try {
+            connection = SnowflakeRuntimeHelper.createConnection(container, rowProperties.getConnectionProperties());
+            if (rowProperties.usePreparedStatement.getValue()) {
+                try (PreparedStatement pstmt = connection.prepareStatement(rowProperties.query.getValue())) {
+                    SnowflakePreparedStatementUtils.fillPreparedStatement(pstmt, rowProperties.preparedStatementTable);
+                    pstmt.execute();
+                }
+            } else {
```
gonchar-ivan: Similar code is used in 3 classes:
SnowflakeRowWriter
SnowflakeRowReader
SnowflakeRowStandalone

1) This is code is duplicated
2) The same if-else is used each time

Please, consider to make a Strategy class out of this code:
1. analyze abstract steps this code does (but in different manner in if and else parts)
these steps are:
create statement
execute statement
close statement
2. So, we see there are steps, but they are done differently depending is ts Statement or PreparedStatement
3. Create interface (abstract class) (Abstract)StatementExecutor with methods listed above. Seems, it should abstract class, because in both cases close statement action is the same. (Hint, it may extend/implement AutoCloseable)
4. Implement StatementExecutor and PreparedStatmentExecutor classes 
5. Modify Runtime classes constructors to accept/create StatementExecutor
6. Replace the code with calls to StatementExecutor

gonchar-ivan: And after this refactoring you will be able to move StatementExecutor hierarchy in components-common project and allow users to override fillPreparedStatement method. Thus, we'll avoid duplication in SnowflakeRow and JdbcRow

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+ */
+public class SnowflakeRowWriter implements WriterWithFeedback<Result, IndexedRecord, IndexedRecord> {
+
+    private transient static final Logger LOGGER = LoggerFactory.getLogger(SnowflakeRowWriter.class);
+
+    private static final I18nMessages I18N_MESSAGES = GlobalI18N.getI18nMessageProvider()
+            .getI18nMessages(SnowflakeRowWriter.class);
+
+    private static final Set<String> CUD_RESULT_SET_COLUMN_NAMES = new HashSet<>(
+            Arrays.asList(new String[] { "number of rows inserted", "number of rows updated", "number of rows deleted" }));
```
gonchar-ivan: What does "CUD" means?
Seems, it worse to add javadoc

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+
+    private final WriteOperation<Result> writeOperation;
+
+    private final SnowflakeRowSink sink;
+
+    private final TSnowflakeRowProperties rowProperties;
+
+    private Connection connection;
+
+    private int commitCount;
```
gonchar-ivan: doesn't it correspond to totalCount?

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    private void handleSuccess() throws SQLException {
+
+        Schema outputSchema = rowProperties.schemaFlow.schema.getValue();
+        if (outputSchema == null || outputSchema.getFields().size() == 0) {
+            return;
+        }
+
+        if (rs == null || (!resultSetValidation && !validateResultSet())) {
```
gonchar-ivan: is it possible to move `!resultSetValidation` check inside validateResultSet() method?

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+    private transient JDBCResultSetIndexedRecordConverter resultSetFactory;
+
+    // Shows if Result Set is compatible with component Schema
+    private transient boolean resultSetValidation = false;
+
+    private transient Schema mainSchema;
+
+    private Result result;
+
+    public SnowflakeRowWriter(RuntimeContainer adaptor, SnowflakeRowWriteOperation writeOperation) {
```
gonchar-ivan: please rename adaptor -> container

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+        this.container = adaptor;
+        this.writeOperation = writeOperation;
+        this.sink = writeOperation.getSink();
+        this.rowProperties = sink.getRowProperties();
+    }
+
+    @Override
+    public void open(String uId) throws IOException {
+        successfulWrites = new ArrayList<>();
```
gonchar-ivan: should it be cleared sometimes?

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+        rejectedWrites = new ArrayList<>();
+
+        connection = sink.connect(container);
+
+        mainSchema = rowProperties.table.main.schema.getValue();
+        if (AvroUtils.isIncludeAllFields(mainSchema)) {
+            mainSchema = sink.getEndpointSchema(container, rowProperties.table.tableName.getValue());
+        }
```
gonchar-ivan: please, see my comment about accessing properties

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+
+        try {
+            if (rowProperties.usePreparedStatement.getValue()) {
+                statement = connection.prepareStatement(rowProperties.query.getValue());
+            } else {
+                statement = connection.createStatement();
+            }
+        } catch (SQLException e) {
+            throw new IOException(e);
+        }
```
gonchar-ivan: please, consider to use StatementExecutor here

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+        try {
+            if (rowProperties.usePreparedStatement.getValue()) {
+                PreparedStatement pstmt = (PreparedStatement) statement;
+                SnowflakePreparedStatementUtils.fillPreparedStatement(pstmt, rowProperties.preparedStatementTable);
+                pstmt.execute();
+                rs = pstmt.getResultSet();
+                pstmt.clearParameters();
+            } else {
+                rs = statement.executeQuery(rowProperties.query.getValue());
+            }
```
gonchar-ivan: StatementExecutor could be used here. It will be something like
```
try {
   rs = statementExecutor.execute();
   handleSuccess();
} catch ...
```

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+
+            // We should return the result of query execution, instead of returning incoming value.
+            handleSuccess();
+        } catch (SQLException e) {
+            if (rowProperties.dieOnError.getValue()) {
+                throw new IOException(e);
+            }
+            LOGGER.error(I18N_MESSAGES.getMessage("error.queryExecution"), e);
```
gonchar-ivan: the logic is similar to exception handling in SnowflakeRowStandalone
Maybe it is worse to extract this code to common method for both classes (ExceptionHandler.handleException(String messageCode) throws IOException )

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+                    && commitCount == rowProperties.commitCount.getValue()) {
+                connection.commit();
+                commitCount = 0;
+            }
+        } catch (SQLException e) {
+            if (rowProperties.dieOnError.getValue()) {
+                throw new IOException(e);
+            }
+            LOGGER.error(I18N_MESSAGES.getMessage("error.performCommit"), e);
+        }
```
gonchar-ivan: I suggest to extract this code to commitIfRequired() method

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+
+        if (rs == null || (!resultSetValidation && !validateResultSet())) {
+            return;
+        }
+
+        if (resultSetFactory == null) {
+            resultSetFactory = new SnowflakeResultSetIndexedRecordConverter();
+            resultSetFactory.setSchema(mainSchema);
+        }
```
gonchar-ivan: please, extract this code to `getConverter()` method

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+
+        while (rs.next()) {
+            result.totalCount++;
+            result.successCount++;
+            IndexedRecord resultSetIndexedRecord = resultSetFactory.convertToAvro(rs);
```
gonchar-ivan: can be replaced with
`getConverter().convertToAvro(rs)`

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/SnowflakeRowWriter.java
```diff
@@ -0,0 +1,300 @@
...
+                successfulWrites.add(resultSetIndexedRecord);
+            } else {
+                IndexedRecord output = new GenericData.Record(outputSchema);
+                // On this moment schemas will be the same, since schema validation has passed.
+                for (Field outField : outputSchema.getFields()) {
+                    Field inputField = resultSetIndexedRecord.getSchema().getField(outField.name());
+                    if (inputField != null) {
+                        output.put(outField.pos(), resultSetIndexedRecord.get(inputField.pos()));
+                    }
+                }
```
gonchar-ivan: Can't you do this action inside converter.convertToAvro()?
Note, you may pass both schemas to Converter constructor or you may have 2 Converters (or 2 implementations, but you may choose which implementation to instantiate according schema)

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/utils/SnowflakePreparedStatementUtils.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Date;
+import java.util.List;
+
+import org.talend.components.snowflake.SnowflakePreparedStatementTableProperties;
+
+/**
+ * Utility class that helps inserting values into prepared statement.
+ *
+ */
+public final class SnowflakePreparedStatementUtils {
```
gonchar-ivan: If you will implement StatementExecutor, then this code can be moved to PreparedStatementExecutor class (and this class removed)

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/utils/SnowflakeRuntimeHelper.java
```diff
@@ -0,0 +1,115 @@
...
+import org.talend.components.api.container.RuntimeContainer;
+import org.talend.components.snowflake.SnowflakeConnectionProperties;
+import org.talend.components.snowflake.runtime.SnowflakeConstants;
+import org.talend.daikon.i18n.GlobalI18N;
+import org.talend.daikon.i18n.I18nMessages;
+
+/**
+ * Contains methods for creating and closing Connection.
+ */
+public class SnowflakeRuntimeHelper {
```
gonchar-ivan: All code from this class may go to SnowflakeRuntime parent class

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowDefinition.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public Class<? extends ComponentProperties> getPropertyClass() {
+        return TSnowflakeRowProperties.class;
+    }
+
+    @Override
+    public boolean isConditionalInputs() {
+        return true;
+    }
```
gonchar-ivan: do you know meaning of this method. Maybe it's not required

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowDefinition.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public String getPartitioning() {
+        return AUTO;
+    }
```
gonchar-ivan: do you know meaning of this method. Maybe it's not required

mbasiuk-talend: Other Rows are using this. So I suggest that my implementation should have the same.

components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakerow/TSnowflakeRowDefinition.java
mbasiuk-talend: Other rows have the same conditional value

components/components-snowflake/src/main/java/org/talend/components/snowflake/runtime/utils/SnowflakeRuntimeHelper.java
mbasiuk-talend: Agree

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1097/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1097/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/1101/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/1101/) 
[Sonar results](https://sonar.datapwn.com/overview?id=org.talend.components%3Acomponents%3Ambasiuk%2FTDI-39295_snowflake_row)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/1103/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/1103/)

mbasiuk-talend:
Latest build failed due to aborting it by iwangtalend. Previous build was successful and after small refactoring tests locally passed.

PR #1153, opened 2018-02-13
feat(TDI-39983): Case sensitivity SnowflakeOutput

olhava:
* added migration task using TComp approach

**What is the new behavior?**

- [ ] Bugfix
- [x] Feature
- [x] Refactoring (no functional changes, no api changes)

- [ ] Yes
- [x] No

**Other information**:


components/components-snowflake/src/main/java/org/talend/components/snowflake/tsnowflakeoutput/TSnowflakeOutputProperties.java
```diff
@@ -215,5 +217,18 @@ public void afterMainSchema() {
...
 
+    @Override
+    public int getVersionNumber() {
+        return 1;
+    }
 
+    @Override
+    public boolean postDeserialize(int version, PostDeserializeSetup setup, boolean persistent) {
+        boolean migrated = super.postDeserialize(version, setup, persistent);
+        if (version < this.getVersionNumber()) {
```
gonchar-ivan: Potentially, component will have multiple versions and we will implement several migration. Probably, we will want to know which migration is related to which version. Maybe it is better to change if condition in following way:
`version < 1`
thus, it will be clear that if clause body is related to migration for version 1
What do you think?

olhava: I agree with you, I'll also add a named constant for "1" to improve code readability.

PR #1429, opened 2018-11-20
fix(TDI-41291): use global search when no folder found in path

undx:

In `tGoogleDriveList` when using `Access By Name` if the folder is not in the root path, the search returns nothing.

https://jira.talendforge.org/browse/TDI-41291

**What is the new behavior?**
Now, if no folder is found we make a global search on the folder's name.

- [x] The code coverage on new code >75%

- [X] Bugfix
- [ ] Feature

- [ ] Yes
- [X] No

**Other information**:


components/components-googledrive/components-googledrive-runtime/src/main/java/org/talend/components/google/drive/runtime/GoogleDriveUtils.java
```diff
@@ -149,6 +149,12 @@ public String getResourceId(String query, String resourceName, String type) thro
         List<String> path = getExplodedPath(folderName);
         String start = path.get(0);
         result = checkPath(0, path, start, "root", searchInTrash);
+        LOG.debug("[getFolderIds] checkPath returned : {}", result);
+        if (result.isEmpty() && !folderName.contains(PATH_SEPARATOR)){
```
ypiel-talend: what is "!folderName.contains(PATH_SEPARATOR)" for ?

undx: We cannot use global search with a pseudo path. So skipping when we're in that case.

PR #1539, opened 2019-10-14
feat(TDI-42911) Migration after Remove CryptoHelper to use AES

bhe-talendbj:
https://jira.talendforge.org/browse/TDI-42911

This PR is depend on daikon PR

https://github.com/Talend/daikon/pull/558

**What is the new behavior?**
Encrypt password field with AES encryption algorithm instead of DES

- [x] The new code does not introduce new technical issues (sonar / eslint)

- [ ] Bugfix
- [x] Feature

- [ ] Yes
- [x] No

**Other information**:

Changes included in this PR

- StringProperty(if ENCRYPT flag is set) will be encrypted by AES encryption

- Decrypt StringProperty(if ENCRYPT flag is set) by AES encryption or CryptoHelper depending on the encrypted value.

Also the component.encryption.key is not supposed to be updated by customer, since this encryption key will be required in

- studio

- runtime of generated job


build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/97/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/97/)

bhe-talendbj:
> This PR is to replace the previous "routine" encryption key?

No. This PR is to remove CryptoHelper from studio components which depend on StringProperty, StringProperty depends on CryptoHelper.

Component properties will be serialized to disk as items and packaged into generated job and  loaded into studio.

While serializing component properties, StringProperty(if ENCRYPT flag is set) will be encrypted by CryptoHelper.

This PR is to 

1) StringProperty(if ENCRYPT flag is set) will be encrypted by AES encryption
2) Decrypt StringProperty(if ENCRYPT flag is set) by AES encryption or CryptoHelper depending on the encrypted value.

Also the component.encryption.key is not supposed to be updated by customer, since this encryption key will be required in

1) studio
2) generated job

components-parent/components-bom/pom.xml
```diff
@@ -10,7 +10,7 @@
     <name>Component BOM</name>
     <properties>
         <components.version>0.28.0-SNAPSHOT</components.version>
-        <daikon.version>0.31.8</daikon.version>
```
wwang-talend: here we change the daikon version, also need to update the daikon version in studio, here only affect the one in talend m2 repository plugin : maven.tos/tis

core/components-api/src/main/java/org/talend/components/api/properties/ComponentBasePropertiesImpl.java
```diff
@@ -0,0 +1,104 @@
...
+import org.talend.daikon.properties.property.Property.Flags;
+import org.talend.daikon.properties.property.StringProperty;
+import org.talend.daikon.serialize.PostDeserializeSetup;
+
+/*
+ * <p>The purpose of this class is to encrypt and decrypt component's password fields with AES algorithm</p> <p>
+ * <ul>Decrypt password field with AES or DES algorithm is based on the value of password field</ul> <ul>Encrypt
+ * password field with AES algorithm always from now on, since DES is deprecated</ul> </p> Created by bhe on Oct 8, 2019
+ */
+public class ComponentBasePropertiesImpl extends PropertiesImpl {
```
wwang-talend: the properties don't only work for studio, also works for other platform like dataprep, though it will be replaced by tacokit. do you consider that?

components-parent/components-bom/pom.xml
wwang-talend: and if you change the daikon version, it is not a little task, as even some TDQ component use daikon too.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/103/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/103/)

components-parent/components-bom/pom.xml
bhe-talendbj: 0.31.10 should be compatible with 0.31.8, daikon of studio was upgraded to 0.31.10 too.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/104/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/104/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/105/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/105/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/106/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/106/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/107/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/107/)

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
```diff
@@ -0,0 +1,147 @@
...
+            if (ks.getKey() != null) {
+                return ks;
+            }
+        } catch (Exception e) {
+            LOGGER.log(Level.FINEST, "Can not load key: " + encryptionKeyName + " from system properties");
+        }
+        return null;
+    }
+
+    private static class ResourceKeyFileSource implements KeySource {
```
coheigea: I think it's better to make ResourceKeyFileSource a standalone class and share it with StudioEncryption, making it configurable so that it can be re-used by both.

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+
+        @Override
+        public byte[] getKey() throws Exception {
+            String key = p.getProperty(KEY_COMPONENT);
+            if (key == null) {
+                throw new IllegalArgumentException("Invalid key name");
+            } else {
+                byte[] keyData = Base64.getDecoder().decode(key.getBytes(StandardCharsets.UTF_8));
```
coheigea: No need to declare the array, just return it.

core/components-api/src/main/java/org/talend/components/api/properties/ComponentBasePropertiesImpl.java
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * @param name
+     */
+    public ComponentBasePropertiesImpl(String name) {
+        super(name);
+    }
+
+    @Override
+    public boolean postDeserialize(int version, PostDeserializeSetup setup, boolean persistent) {
```
coheigea: It's not clear to my why we are overriding the Daikon class, when instead we could be making the changes required in Daikon StringProperty.encryptStoredValue? That way every class that picks up StringProperty would see the changes....

bhe-talendbj: If we change StringProperty, need to inject Encryption object into StringProperty based on crypto-utils, so that StringProperty.encryptStoredValue can depend on that Encryption object. However Encryption object can not be serialized to disk. So after StringProperty is loaded from disk, StringProperty can not construct Encryption object.

So I have to override the daikon class.

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
coheigea: No it's OK then

core/components-api/src/main/java/org/talend/components/api/properties/ComponentBasePropertiesImpl.java
coheigea: I don't understand though why we would need to serialize the Encryption object in StringProperty though. We could just change the encryptStoredValue method to take a Function<String, String> and use this to do the encryption. Or we could take the encryption out altogether and put it in PropertiesImpl instead.

bhe-talendbj: 1) If we make changes like encryptStoredValue(Function<String, String> encrypt, Function<String, String> decrypt), We need to change existing interfaces of PropertiesImpl or add new interfaces to PropertiesImpl. encryptStoredValue is called by PropertiesImpl. Then we need to set two functions to PropertiesImpl or its subclass.

2) >>Or we could take the encryption out altogether and put it in PropertiesImpl instead.

We can add following method to PropertiesImpl

public void setEncryptAndDecryptFuncs(Function<String, String> encrypt, Function<String, String> decrypt)

Can I implement 2)?

coheigea: Yes that would be fine. It should default to the old CryptoHelper. Let me know when you have the Daikon PR ready and I can review that.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/108/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/108/)

core/components-api/src/main/java/org/talend/components/api/properties/ComponentBasePropertiesImpl.java
coheigea: We don't need to override these two methods any more - just encryptData - right?

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/112/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/112/)

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/113/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/113/)

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
```diff
@@ -0,0 +1,106 @@
...
+/*
+ * Created by bhe on Sep 27, 2019
+ */
+public class ComponentEncryption {
+
+    private static final String KEY_COMPONENT_DEFAULT = ComponentKeySource.KEY_SYSTEM_PREFIX + "1";
+
+    private static final String ENCRYPTION_PREFIX = "enc:";
+
+    private static final Pattern REG_ENCRYPTED_DATA = Pattern.compile("^enc\\:system\\.encryption\\.key\\.v\\d\\:.+");
```
coheigea: Instead of .+ at the end, maybe we can use the BASE-64 encoded alphabet?

core/components-api/src/main/java/org/talend/components/api/properties/ComponentBasePropertiesImpl.java
```diff
@@ -0,0 +1,52 @@
...
+        super(name);
+    }
+
+    @Override
+    protected void encryptData(Property property, boolean encrypt) {
+        if (property.getStoredValue() == null) {
+            return;
+        }
+
+        String oldValue = String.valueOf(property.getStoredValue());
```
coheigea: rename "oldValue" here as it's a bit misleading. Just storedValue would be fine.

core/components-api/src/main/java/org/talend/components/api/properties/ComponentKeySource.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private static int getVersion(String keyName) {
+        String[] keyNameArray = keyName.split("\\.");
+        if (keyNameArray[keyNameArray.length - 1].startsWith("v")) {
+            try {
+                return Integer.parseInt(keyNameArray[keyNameArray.length - 1].substring(1));
+            } catch (NumberFormatException e) {
+                LOGGER.log(Level.WARNING, "Parse version of encryption key error, key: " + keyName);
+            }
+        }
```
coheigea: I don't think it's efficient to call "split" here, as all you care about is the last segment. You could instead call keyName.lastIndexOf('.') and then call substring using the index

core/components-api/src/main/java/org/talend/components/api/properties/ComponentKeySource.java
```diff
@@ -0,0 +1,157 @@
...
+                    keyProperties.load(fi);
+                } catch (IOException e) {
+                    LOGGER.log(Level.SEVERE, e.getMessage(), e);
+                }
+            }
+        }
+
+        // load system key data from System properties
+        System.getProperties().forEach((k, v) -> {
+            String key = String.valueOf(v);
```
coheigea: Should this not be String.valueOf(k) here?

core/components-api/src/main/java/org/talend/components/api/properties/ComponentKeySource.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(ComponentKeySource.class.getName());
+
+    private static final String ENCRYPTION_KEY_FILE_SYS_PROP = "encryption.keys.file";
+
+    public static final String KEY_SYSTEM_PREFIX = "system.encryption.key.v";
+
+    private static final String DEFAULT_KEY_FILE_NAME = "component.keys";
+
+    private final String requestKey;
```
coheigea: Rename this - "requestKey" sounds like it is the key itself, but it's only the key name

core/components-api/src/main/java/org/talend/components/api/properties/ComponentKeySource.java
```diff
@@ -0,0 +1,157 @@
...
+        int keyVersion = getVersion(this.requestKey);
+        // No version
+        if (keyVersion == 0) {
+            return this.requestKey;
+        }
+
+        String keyPrefix = getKeyPrefix(this.requestKey);
+
+        return this.availableKeys.stringPropertyNames().stream().filter(e -> e.startsWith(keyPrefix))
+                .max(Comparator.comparing(e -> getVersion(e))).get();
```
coheigea: Please explain the logic here in a comment - it took me a while to figure it out

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Encrypt input data
+     */
+    public static String encrypt(String src) {
+        if (src == null || isEncrypted(src)) {
+            return src;
+        }
+        try {
+            ComponentKeySource ks = getKeySource(KEY_COMPONENT_DEFAULT, true);
+            return ENCRYPTION_PREFIX + ks.getKeyName() + ":" + new Encryption(ks, CipherSources.getDefault()).encrypt(src);
```
coheigea: Use StringBuilder here.

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
```diff
@@ -0,0 +1,106 @@
...
+
+import org.talend.daikon.crypto.CipherSources;
+import org.talend.daikon.crypto.Encryption;
+
+/*
+ * Created by bhe on Sep 27, 2019
+ */
+public class ComponentEncryption {
+
+    private static final String KEY_COMPONENT_DEFAULT = ComponentKeySource.KEY_SYSTEM_PREFIX + "1";
```
coheigea: I think we should get rid of this and change the logic in ComponentKeySource. For encryption, we should just pass the system prefix in as the key name, and have it select the highest key version number that corresponds to the prefix.

core/components-api/src/main/java/org/talend/components/api/properties/ComponentKeySource.java
```diff
@@ -0,0 +1,157 @@
...
+
+        int keyVersion = getVersion(this.requestKey);
+        // No version
+        if (keyVersion == 0) {
+            return this.requestKey;
+        }
+
+        String keyPrefix = getKeyPrefix(this.requestKey);
+
+        return this.availableKeys.stringPropertyNames().stream().filter(e -> e.startsWith(keyPrefix))
```
coheigea: For encryption, the requestKey is always "system.encryption.key.v" right? In that case, we could just do this filter once after creating the Properties object, and store the highest name of the highest "v" key. Then each time we call this method for encryption we don't have to perform the stream.

components/components-fileio/simplefileio-definition/src/main/java/org/talend/components/simplefileio/s3/S3DatastoreProperties.java
```diff
@@ -17,14 +17,14 @@
...
+import org.talend.components.api.properties.ComponentBasePropertiesImpl;
 import org.talend.components.common.datastore.DatastoreProperties;
-import org.talend.daikon.properties.PropertiesImpl;
 import org.talend.daikon.properties.presentation.Form;
 import org.talend.daikon.properties.presentation.Widget;
 import org.talend.daikon.properties.property.Property;
 import org.talend.daikon.properties.property.PropertyFactory;
 
-public class S3DatastoreProperties extends PropertiesImpl implements DatastoreProperties {
+public class S3DatastoreProperties extends ComponentBasePropertiesImpl implements DatastoreProperties {
```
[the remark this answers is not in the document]
bhe-talendbj: It is only necessary to make changes to those properties which depends on StringProperty and set ENCRYPT flag.

I made changes to above properties found only.

core/components-api/src/main/java/org/talend/components/api/properties/ComponentBasePropertiesImpl.java
```diff
@@ -0,0 +1,52 @@
...
+            // always encrypt data by AES
+            property.setStoredValue(ComponentEncryption.encrypt(oldValue));
+        } else {
+            if (ComponentEncryption.isEncrypted(oldValue)) {
+                property.setStoredValue(ComponentEncryption.decrypt(oldValue));
+                return;
+            }
+            // fallback to CryptoHelper
+            super.encryptData(property, encrypt);
+        }
```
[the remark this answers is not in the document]
bhe-talendbj: Done

core/components-api/src/main/java/org/talend/components/api/properties/ComponentBasePropertiesImpl.java
bhe-talendbj: Done.

core/components-api/src/main/java/org/talend/components/api/properties/ComponentKeySource.java
bhe-talendbj: Done

core/components-api/src/main/java/org/talend/components/api/properties/ComponentKeySource.java
bhe-talendbj: Done.

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
bhe-talendbj: Done

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
bhe-talendbj: Done

core/components-api/src/main/java/org/talend/components/api/properties/ComponentKeySource.java
bhe-talendbj: Done.

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/114/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/114/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/115/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/115/) 
[Sonar results](https://sonar.datapwn.com/overview?id=org.talend.components%3Acomponents%3Abhe%2Ffeat%2FTDI-42911_Remove_CryptoHelper)

build-talend-common:
[Sonar results](https://sonar.datapwn.com/overview?id=org.talend.components%3Acomponents%3Abhe%2Ffeat%2FTDI-42911_Remove_CryptoHelper)

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
```diff
@@ -0,0 +1,117 @@
...
+    private static final Logger LOGGER = Logger.getLogger(ComponentEncryption.class.getCanonicalName());
+
+    private static final ThreadLocal<Properties> LOCALCACHEDALLKEYS = ThreadLocal.withInitial(() -> {
+        return ComponentKeySource.loadAllKeys();
+    });
+
+    // cached key source for encrytion
+    private static final ThreadLocal<ComponentKeySource> LOCALCACHEDKEYFORENCRYPT = ThreadLocal.withInitial(() -> {
+        return ComponentKeySource.key(ComponentKeySource.loadAllKeys(), ComponentKeySource.KEY_SYSTEM_PREFIX, true);
+    });
```
coheigea: I don't understand why we need two ThreadLocals here, one for encryption and one for decryption? I think we should just have one. For decryption we ask for the key with the given name. For encryption, it selects the highest "v" key for the prefix.

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
```diff
@@ -0,0 +1,117 @@
...
+import org.talend.daikon.crypto.Encryption;
+
+/*
+ * Created by bhe on Sep 27, 2019
+ */
+public class ComponentEncryption {
+
+    private static final String ENCRYPTION_PREFIX = "enc:";
+
+    private static final Pattern REG_ENCRYPTED_DATA = Pattern.compile("^enc\\:system\\.encryption\\.key\\.v\\d\\:\\p{Print}+");
```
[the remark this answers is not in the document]
bhe-talendbj: Done

core/components-api/src/main/java/org/talend/components/api/properties/ComponentEncryption.java
bhe-talendbj: Done

build-talend-common:
Pull request test result: [failure](https://ci-common.datapwn.com/job/components_pull-request_master/117/) [:x:](https://ci-common.datapwn.com/job/components_pull-request_master/117/)

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/120/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/120/) 
[Sonar results](https://sonar.datapwn.com/overview?id=org.talend.components%3Acomponents%3Abhe%2Ffeat%2FTDI-42911_Remove_CryptoHelper)

core/components-api/src/main/java/org/talend/components/api/properties/ComponentKeySource.java
```diff
@@ -0,0 +1,153 @@
...
+
+    private final Properties availableKeys;
+
+    private String keyForEncrypt;
+
+    private ComponentKeySource(Properties allKeys, String keyName, boolean isEncrypt) {
+        this.availableKeys = allKeys;
+        this.keyName = keyName;
+        this.isEncrypt = isEncrypt;
+        this.keyForEncrypt = this.availableKeys.stringPropertyNames().stream().max(Comparator.comparing(e -> getVersion(e)))
```
coheigea: I don't think we need a separate property for encryption. Instead it could be something like:
if (isEncrypt) {
    this.keyName = this.availableKeys....
} else {
    this.KeyName = keyName;
}

bhe-talendbj: If I made the changes you suggested here, then key for encryption will be calculated(stream and compare) for every calls of getKey(), is this expected?

>> coheigea 9 days ago Member
For encryption, the requestKey is always "system.encryption.key.v" right? In that case, we could just do this filter once after creating the Properties object, and store the highest name of the highest "v" key. Then each time we call this method for encryption we don't have to perform the stream.

It seems that your above previous comment is more reasonable?

build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/125/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/125/) 
[Sonar results](https://sonar.datapwn.com/overview?id=org.talend.components%3Acomponents%3Abhe%2Ffeat%2FTDI-42911_Remove_CryptoHelper)

PR #1924, opened 2021-05-21
TDI-46133 - Switch Beam GeneratorFunction to use SHA-256 instead of MD5

coheigea:
This task is to switch the Beam GeneratorFunction to use SHA-256 instead of MD5. It is only being used for test-code, but we might as well eliminate MD5 as it is a deprecated algorithm.


build-talend-common:
Pull request test result: [success](https://ci-common.datapwn.com/job/components_pull-request_master/807/) [:star:](https://ci-common.datapwn.com/job/components_pull-request_master/807/)

wwang-talend (approved):

coheigea:
@wwang-talend Please merge if convenient

PR #2038, opened 2022-03-17
fix(TDI-47651): add redshift driver class

ozhelezniak-talend:
https://jira.talendforge.org/browse/TDI-47651

**What is the new behavior?**
added a new jdbc redshift driver class


undx (approved):
LGTM

PR #2042, opened 2022-03-22
fix(TDI-47168): Add avro name validation

dmytro-sylaiev:

**What is the new behavior?**

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:


clesaec (commented):
given the stacktrace inside log file, shouldn't it be better to fix issue in [MetadataToolAvroHelper](https://github.com/Talend/tcommon-studio-se/blob/master/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java#L62) class ? 
(it will be more general, and you even can store the original name in Avrp properties schema, here, you have no method to back to original name, can't know if _123 is original or come from 123).
[AvroNamesValidationHelper](https://github.com/Talend/daikon/blob/master/daikon/src/main/java/org/talend/daikon/avro/AvroNamesValidationHelper.java#L41) just add underscore, assume that input name is Java name. In Snowflake case, is it the case ?

PR #2058, opened 2022-04-05
fix(TDI-47772): Snowflake CHAR precision error when create table action

dochkas-talend:
Added CHAR type to Snowflake types


dmytrogrygorenko (approved):
LGTM

PR #2103, opened 2022-07-14
fix(TDI-48093): Allow GoogleDrive component to work with large files.

dmytrogrygorenko:
https://jira.talendforge.org/browse/TDI-48093

**What is the new behavior?**

- [X] Bugfix
- [ ] Feature

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/components/components-googledrive/components-googledrive-definition/src/main/java/org/talend/components/google/drive/get/GoogleDriveGetDefinition.java
+++ b/components/components-googledrive/components-googledrive-definition/src/main/java/org/talend/components/google/drive/get/GoogleDriveGetDefinition.java
@@ -46,13 +46,13 @@ public class GoogleDriveGetDefinition extends GoogleDriveComponentDefinition {
             ConnectorTopology connectorTopology) {
         assertEngineCompatibility(engine);
         assertConnectorTopologyCompatibility(connectorTopology);
-		switch (connectorTopology) {
-		case NONE:
-			return getRuntimeInfo(GoogleDriveConnectionDefinition.GET_RUNTIME_CLASS);
-		case OUTGOING:
-			return getRuntimeInfo(GoogleDriveConnectionDefinition.SOURCE_CLASS);
-		}
-		return null;
+        switch (connectorTopology) {
+        case NONE:
+            return getRuntimeInfo(GoogleDriveConnectionDefinition.GET_RUNTIME_CLASS);
+        case OUTGOING:
+            return getRuntimeInfo(GoogleDriveConnectionDefinition.SOURCE_CLASS);
+        }
+        return null;
     }
 
     @Override
--- a/components/components-googledrive/components-googledrive-definition/src/main/java/org/talend/components/google/drive/get/GoogleDriveGetDefinition.java
+++ b/components/components-googledrive/components-googledrive-definition/src/main/java/org/talend/components/google/drive/get/GoogleDriveGetDefinition.java
@@ -46,13 +46,13 @@ public class GoogleDriveGetDefinition extends GoogleDriveComponentDefinition {
             ConnectorTopology connectorTopology) {
         assertEngineCompatibility(engine);
         assertConnectorTopologyCompatibility(connectorTopology);
-        if (ConnectorTopology.NONE.equals(connectorTopology)) {
-            return getRuntimeInfo(GoogleDriveConnectionDefinition.GET_RUNTIME_CLASS);
-        } else if (ConnectorTopology.OUTGOING.equals(connectorTopology)) {
-            return getRuntimeInfo(GoogleDriveConnectionDefinition.SOURCE_CLASS);
-        } else {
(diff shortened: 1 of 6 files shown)


components/components-googledrive/components-googledrive-runtime/src/main/java/org/talend/components/google/drive/runtime/GoogleDriveUtils.java
```diff
@@ -482,24 +483,40 @@ public GoogleDriveGetResult getResource(GoogleDriveGetParameters parameters) thr
...
+
+        if (!parameters.isCreateByteArray() && parameters.isStoreToLocal()) {
             LOG.info(messages.getMessage("message.writing.resource", parameters.getResourceId(), localFile));
-            try (FileOutputStream fout = new FileOutputStream(localFile)) {
-                fout.write(content);
+        }
+
+        byte[] content = null;
+        try (OutputStream outputStream = (parameters.isCreateByteArray()
+                || (!parameters.isCreateByteArray() && !parameters.isStoreToLocal())) ? new ByteArrayOutputStream()
```
ozhelezniak-talend, line 501: a || (!a && !b) = a || !(a || b)

0-0|1
0-1|0
1-0|1
1-1|1

a || !b

0-0|1
0-1|0
1-0|1
1-1|1

I think you can simplify the condition to 
parameters.isCreateByteArray() || !parameters.isStoreToLocal()

components/components-googledrive/components-googledrive-runtime/src/main/java/org/talend/components/google/drive/runtime/GoogleDriveUtils.java
```diff
@@ -482,24 +483,40 @@ public GoogleDriveGetResult getResource(GoogleDriveGetParameters parameters) thr
...
+        String localFile = null;
         if (parameters.isStoreToLocal()) {
-            String localFile = parameters.getOutputFileName();
+            localFile = parameters.getOutputFileName();
             if (parameters.isAddExt()) {
                 localFile = localFile + ((localFile.endsWith(outputFileExt)) ? "" : outputFileExt);
             }
+        }
+
+        if (!parameters.isCreateByteArray() && parameters.isStoreToLocal()) {
```
ozhelezniak-talend, line 495: remain isStoreToLocal
and you don't need to duplicate this message on line 515

ozhelezniak-talend (approved):
Looks lgtm.
The logic of file download becomes complicated. It's good to refactor it a bit.
Whitespaces conflict: tab vs spaces. Looks not homogeneous.

components/components-googledrive/components-googledrive-definition/src/main/java/org/talend/components/google/drive/get/GoogleDriveGetDefinition.java
```diff
@@ -46,13 +46,13 @@ public RuntimeInfo getRuntimeInfo(ExecutionEngine engine, ComponentProperties pr
...
         assertEngineCompatibility(engine);
         assertConnectorTopologyCompatibility(connectorTopology);
-        if (ConnectorTopology.NONE.equals(connectorTopology)) {
-            return getRuntimeInfo(GoogleDriveConnectionDefinition.GET_RUNTIME_CLASS);
-        } else if (ConnectorTopology.OUTGOING.equals(connectorTopology)) {
-            return getRuntimeInfo(GoogleDriveConnectionDefinition.SOURCE_CLASS);
-        } else {
-            return null;
-        }
+		switch (connectorTopology) {
```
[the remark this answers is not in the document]
dmytrogrygorenko: Fixed.

components/components-googledrive/components-googledrive-runtime/src/main/java/org/talend/components/google/drive/runtime/GoogleDriveUtils.java
dmytrogrygorenko: Fixed. Thank you for the suggestion!

components/components-googledrive/components-googledrive-runtime/src/main/java/org/talend/components/google/drive/runtime/GoogleDriveUtils.java
dmytrogrygorenko: It can be leaved as it is, IMO. If we use byte array - we may not use "Local File" option, but if we do, creating of a byte array will take some time, depending on the file size, and this log message will fire only when byte array is successfully created, which can be useful during debug. If we don't use byte array, this message should be fired right before we start using an output stream to the file. So basically, there are two separate places in the code where we should log this output action.

ozhelezniak-talend (approved):
lgtm (I'd prefer a refactoring because logic becomes complicated)
but it should work

PR #2131, opened 2022-09-23
fix(TDI-48546): tSalesforceOutput component throws java.lang.StringIndexOutOfBoundsException when the Parallelization is enabled

wwang-talend:
https://jira.talendforge.org/browse/TDI-48546

**What is the new behavior?**
https://jira.talendforge.org/browse/TDI-48546

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

### DIFF
--- a/components/components-salesforce/components-salesforce-runtime/src/main/java/org/talend/components/salesforce/connection/oauth/SalesforceImplicitConnection.java
+++ b/components/components-salesforce/components-salesforce-runtime/src/main/java/org/talend/components/salesforce/connection/oauth/SalesforceImplicitConnection.java
@@ -33,7 +33,7 @@ public class SalesforceImplicitConnection {
 
     private final Logger LOGGER = LoggerFactory.getLogger(SalesforceImplicitConnection.class.getName());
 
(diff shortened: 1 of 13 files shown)


vyu-talend (approved):
LGTM

wwang-talend:
reason : we share background java.text.MessageFormat object in different threads, that object is not thread safe.

solution 1: not share it

solution 2: share it and add syn when use that object's method, but as the object is static, so global unique, so we will face performance down as we have to use a global unqiue lock to lock it, so not like this.

solution 3: make it thread local, but we have to consider the issue release the thread local cache before thread die self, that's the thread pool case, unlucky, our studio iterate line with parallize is that case, image "tloop(enable parallize on the iterate line with number of works : 3)==>(tfixedflowinput==>tsalesforceoutput)". If without release code, it mean we will keep the MessageFormat objects in memory always before the thread pool down. And sure that release code will make more complex code, in my view, not worth. Or maybe our thread pool size is small(< 10), but if we have if we enable several parallize outside of the job design above, like the logic : "tloop==>tloop==>tloop==>(tfixeflowinput==>tsalesforceouput)", maybe it can make memory leak, we no need to face it.

PR #2316, opened 2023-08-18
chore(TDI-49532): use custom builder

acatoire:
Use custom builder as rootless on scan and build jobs


lxia-talend (approved):
LGTM

ci/Jenkinsfile-scan
```diff
@@ -59,58 +43,185 @@ spec:
         skipStagesAfterUnstable()
     }
 
-    triggers {
```
lxia-talend, line 62: no need trigger now ?

lxia-talend (approved):
LGTM

ci/Jenkinsfile-scan
acatoire: yes triggers are inside the flux config

PR #2330, opened 2023-09-14
feat(TDI-49532) Correct package stage, credential and tests

acatoire:
Correct package stage, add a missing credential and do not skip test for the packaging step


fdesiles (approved):
LGTM

PR #2332, opened 2023-09-15
feat(TDI-49532) Revert the scan job but not the build

acatoire:
Revert the scan job but not the build


fdesiles (approved):
LGTM

